PR #5, opened 2013-10-14
Treat 00 and + as synonyms.

ChrisHooks:
00 is the international Dialing Code for a lot of 
countries


josefinscreentek:
If many countries already implemented the ITU standard, can't it then be a setting to the plugin if you want to be able to convert it by the ITU standard in addition to +? It's fine if the result using getNumber() returns the number in the format as of today, with the + sign. As long as it can be an ITU setting to be able to translate if customers fill out it using the ITU 00 format and you're in that part of the world where it's mostly used.

jackocnr:
But if the user types 001 that could mean the start of a number for many different countries (see wikipedia link above), so how would we know which flag to display? The 00 system only works when you're physically dialling a phone number because in that case you know for sure what country you're dialling from. In this case we can't know that for sure, and even if you wanted to assume it, based on their IP or something (which is not reliable) then to make it work we would need to maintain a catalog of the local dialling rules for every country, which would (A) add a lot of weight to the plugin, (B) be a maintenance nightmare to keep it up-to-date, and (C) it would add a massive amount of complexity to the code which currently makes assumptions about international dial codes starting with a plus character.

If there was an easy and safe way to add support for this then I would be happy to do it, but actually I think it would be a massive job to implement and maintain, it would add a lot of weight to the plugin, and I don't think it would be reliable anyway, so I'm afraid I really don't think it's worth it just to have the flag automatically update.

josefinscreentek:
I just created a "replacement" function that I call in the validate function to replace "00" with "+" before it actually validated the function and I think that will work at least for my case.
```
function replaceITUStandardOnMobileNumber(mobilephone) {
    if (mobilephone.startsWith('00') && mobilephone.length > 2) {
        return '+' + mobilephone.substring(2, mobilephone.length);
    }
    return mobilephone;
}
```
[...]
           **input.value = replaceITUStandardOnMobileNumber(input.value);**
            if (input.value.trim()) {
                if (**iti.isValidNumber()**) {
  [...]

josefinscreentek:
It is based on this example: 
https://intl-tel-input.com/node_modules/intl-tel-input/examples/gen/is-valid-number.html

And then one extra line here:
// on blur: validate
input.addEventListener('blur', function() {
  reset();
**input.value = replaceITUStandardOnMobileNumber(input.value);**
  if (input.value.trim()) {
    if (iti.isValidNumber()) {
      validMsg.classList.remove("hide");
    } else {
      input.classList.add("error");
      var errorCode = iti.getValidationError();
      errorMsg.innerHTML = errorMap[errorCode];
      errorMsg.classList.remove("hide");
    }
  }
});

jackocnr:
Hmm your original issue was about updating the flag. If you want to discuss validation, please can you provide exact steps to reproduce the issue, along with what you would expect to see and what you actually see?

If I go to that validation demo page, and select UK, and enter a valid english number starting with 0044 it actually says it's valid.

josefinscreentek:
It updates the flag as well if you replace it as above during validation.

josefinscreentek:
Sorry, I also manually triggered "blur" event in my code.
I've now changed it to be called in the reset function instead  
```
 var reset = function () {
            input.classList.remove("error");
            hiddenInput.value = "";
            errorMsg.innerHTML = "";
            errorMsgCon.classList.add("hide");
            validMsg.classList.add("hide");
            input.value = replaceITUStandardOnMobileNumber(input.value);
        };
```

josefinscreentek:
The result of this is that if the user starts by writing 00 in the input field, 00 gets replaced by + and then they can continue writing the rest of it

josefinscreentek:
Sorry for spamming the thread.
But the final result is this:

```
function replaceITUStandardOnMobileNumber(mobilephone) {
    if (mobilephone.startsWith('00') && mobilephone.length > 1) {
        return '+' + mobilephone.substring(2, mobilephone.length);
    }
    return mobilephone;
}
```
```
        var reset = function () {
            input.classList.remove("error");
            hiddenInput.value = "";
            errorMsg.innerHTML = "";
            errorMsgCon.classList.add("hide");
            validMsg.classList.add("hide");
            input.value = replaceITUStandardOnMobileNumber(input.value);
        };
```

jackocnr:
> The result of this is that if the user starts by writing 00 in the input field, 00 gets replaced by + and then they can continue writing the rest of it

Ah well I'm actually strongly against changing what the user types in an input in any way, especially while they're still typing. So I wont be adding anything like this to the core plugin. But yeah it would be very easy for you to do this yourself, outside of the plugin code.

But also, this will only work for certain countries. For other countries it will break, e.g. countries that use an international call prefix of 000 or 001 etc (see wikipedia link above).

PR #126, opened 2014-12-24
Add not allowed codes feature

lutaev:


jackocnr:
Thanks for writing this, but this solution would require a lot of manual maintenance. If we're going to go down this road, I would favor the solution I outlined in [this issue](https://github.com/Bluefieldscom/intl-tel-input/issues/109), which automatically creates a dial code map object based on the main country data.

jackocnr:
Update: or we could just use _getDialCode haha.

PR #132, opened 2015-01-09
Added support for use with Bootstrap .form-control and .input-group

idpaterson:
Bootstrap form controls fill the width of their container and input groups allow buttons and static elements to appear alongside form fields with adjacent border radii removed to make everything look like a cohesive element. However, intl-tel-input's use of a wrapping container prevents the input from filling the width and causes the input to be too narrow and the input addons to be too tall. This affects fields both inside .input-group (first) and standalone form controls not inside .input-group (second):
before

After this change, when applied to an input with the .form-control class, the .intl-tel-input wrapper element is decorated with the .form-control class as well. The input still has the .form-control class, so only minor stylesheet modifications were necessary to ensure that the input matches the metrics and styling of its .form-control container. I opted for a solution that is only applied when the form-control class is present, to avoid the possibility of conflicts for other users.
after

This includes basic tests to ensure that the form-control class is added, but does not actually pull in bootstrap to test rendering.

Let me know if you would like me to commit the built files or if you prefer to just wait for the next release.


idpaterson:
Travis answered that question for me, I didn't notice that it tests the distribution files as-committed rather than building them from source.

jackocnr:
Hey man, thanks for putting this together, but I'm very reluctant to put anything in the codebase that is specific to a 3rd party tool like this. A) lots of people wont be using bootstrap, and this bloats the plugin for them, B) it will require maintenance - if bootstrap change anything then it will break, and C) where do we draw the line? we could start adding little extra compatibility rules for every 3rd party tool...

Instead I would suggest we create a wiki page or something that gives advice on how to make the plugin compatible with popular frameworks like bootstrap (e.g. extra CSS rules that may be useful, or JS snippets you can run after initialising the plugin). Is this something you would be interested in contributing?

That said, one feature I have considered adding before is a containerClass option, to allow devs to add extra classes to the .intl-tel-input container - what do you think about that?

idpaterson:
That would work for my use case. I am using this via [international-phone-number](/mareczek/international-phone-number), an angular directive that wraps intl-tel-input, which could be updated to support a `container-class` attribute.

jackocnr:
Ok great, so I'll close this, but if you have time to code the containerClass option that would be great.

PR #146, opened 2015-01-29
Added showPlaceholder option

daztmb:
In some cases you don't need input placeholder. This option allows to turn off it.
(There's placeholder, when input field in empty. Some users may think it's phone number inputted in the field. I know it sounds unlikely, but can happen)


jackocnr:
This is a good idea, thanks. I think `autoPlaceholder` would be a more descriptive name. If you change that, and update the readme, I will gladly accept the pull request. Thanks again.

jackocnr:
Thanks!

Issue #148, opened 2015-02-02
Tweaking base css to allow better alignment within a input-group


gegere:
Allow input field to use `<div class="input-group">` around it. 

Great widget!

I am looking to have the input field nicely line up against a call button. This is the tweaked CSS to have the country drop down list align correctly within the field and have a button bump up against the right side. The problem seems to be with a iOS device the input field is not selectable. I can't reproduce this with a Safari browser as it the input field is selectable. Here is a screenshot using the iOS device simulator, click select inside the input box: http://cl.ly/image/2O070x10373m

Everything looks good with the style but I can't select the input field via a iOS device. Here is the changed CSS.

```
.intl-tel-input {
    display: block;
    position: initial;
}
```

This is what the UI looks like via the initial CSS when trying to use "input-group": http://cl.ly/image/3U1G2U1s3l3M

To test and override my custom CSS in CodePen add the following CSS
```
#phone .intl-tel-input {
    display: inline-block;
    position: relative;
}
```

jackocnr:
If you look at how Bootstrap styles the `<input>` element, it uses `display: table-cell;`, so I tried applying that to the `.intl-tel-input` container and it worked. Example here: http://codepen.io/jackocnr/pen/pvdgrO

gegere:
So the example I had worked regarding the styling. The problem is using iOS, is trying to input data or clicking the flag menu and selecting an option. **You can't.**

This is the issue, and I am after a solution. I will test the rendered page tomorrow. I believe I tested all element block types. I give it a stronger, deeper test.

gegere:
Hello Jack, following back up on this. I am sure you are busy I didn't read this over in detail and thought I was focused on the UI style / interface as an issue.

So the example I had worked looks great but does not function. **The problem is** when using `<div class="input-group">` on iOS it is trying to capture inputs and not allowing a click  to the flag / menu.

If have taken a [screencast](http://cl.ly/1R3Q0e0Y142o) to let you view the experience.

I would be happy to help get to the bottom of this.

jackocnr:
Ok I have reproduced this on an iPhone - if you view the codepen on Safari then you can't open the flag dropdown and you can't even focus the input. If you view it on Chrome (on iOS) it works fine.

If I had to guess I would say this must be a bootstrap issue, but it requires further investigation.

gegere:
[...]

PR #161, opened 2015-02-11
Correct example code when using Bootstrap

gegere:
Fix for [Related:](#148) this tweak will allow the '[tel]' input field to work with a responsive UI.

[Working example](http://codepen.io/gegere/pen/vEexzJ) - I went a different route as I was trying to use the purposed which looks good on all browsers but does not function correctly on iOS Safari:

`.intl-tel-input {display: table-cell;}`

On iOS Safari the following function does not fire, as I can able to test and confirm.
```
_showDropdown: function() {
    alert('hello');
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,7 +261,11 @@ In order to get the automatic country-specific placeholders, simply omit the pla
 **Bootstrap input groups**  
 Simply add this line to get [input groups](http://getbootstrap.com/components/#input-groups) working properly.
 ```css
-.intl-tel-input {display: table-cell;}
+.intl-tel-input {
+  width: 100%;
+  text-align:left;
+  padding-top: 5px;
+}
 ```
 
 


gegere:
Hello @jackocnr I am following up on this pull request fix, hope my documentation was clear enough.

jackocnr:
I'm not sure about this solution. Why the text-align:left? It doesn't seem to make any difference. And why the padding-top:5px - where did that number come from?

And finally, the div.input-group has display:table, and the span.input-group-btn has display:table-cell - I think that it only makes sense to make the div.intl-tel-input have display:table-cell as well. If this causes a bug in safari mobile, then I think we should talk to them about that. I just need to do more testing, to check it's not some clash with my plugin CSS, or the Bootstrap CSS. We need to distil that problem down to a simple test case and submit a bug report to the right place.

gegere:
Let's break it down.

**Why the text-align:left?** This is needed so the list items align to the left. The parent container will need to use `text-align:center;` so the `div.input-group` expands or shrinks in the center of a responsive layout / grid.

The `padding-top: 5px;` is need because there is a small drop between the field and the button without it, ie: http://cl.ly/image/2W0x1x413A1z

If you use `display:table-cell` the dropdown of country goes crazy and actually locks up the iOS Safari browser I was able to produce this over and over again, each time every time, needing to reset the iOS simulator or closing Safari manually on a iPhone 6 running 8.1.3 and a iPhone 4s running 8.0.

Add `alert('hello')` and see that is does not fire with your suggestion.
```
_showDropdown: function() {
    alert('hello');
}
```

You have noted `.intl-tel-input {display: table-cell;}` in your README.md as a way to use this with Bootstrap making it responsive. While it works with many other browsers It does not work with iOS / Safari. I spent a good amount of time testing multiple devices and browser and this was the most simplistic solution.

jackocnr:
I don't have time for a proper response right now, but I've realised the problem is to do with the dropdown arrow. If you re-check my codepen - I now remove the arrow div from the markup after initialising the plugin, and it fixes the bug: http://codepen.io/jackocnr/pen/pvdgrO

gegere:
I have tested this on my local version and can confirm this is moving in the correct direction. You know more about the javascript than I do. Happy to test once you have more time. Thanks for quickly looking into it.

jackocnr:
I'm moving this discussion to a new issue: https://github.com/Bluefieldscom/intl-tel-input/issues/165

PR #175, opened 2015-03-06
Update intlTelInput.js

lucaspfsd:
Update autoCountryLoaded using cookie to save data and not do duplicate request for the same public ip.


jackocnr:
Thanks for your helpful input, but I've realised I need to refactor this whole system (as per my comment in that issue).

Issue #176, opened 2015-03-07
Submit the country in nationalMode?


fedeisas:
Hi! Is there any way to submit the CC when in nationalMode?

Thanks!

jackocnr:
Hi there. A couple of people have asked this recently, so I added something to the Troubleshooting section of the readme - does that answer your question?

If you have any other questions or suggestions, do let me know.

fedeisas:
I went with the hidden input that syncs the dial code whenever the dropdown changes.

I think users prefer to write down their number in nationalMode (it's easier) and I can prepopulate their country (by IP, user prefs, etc.). But, if they change their country, I wanted my backend to notice, so I submit two fields (`phone` and `country_dial_code`).

I think this could make a nice feature in your library, an option like: `submitCountryCode: (bool)` and then when `_selectListItem` function is run, update the hidden input.

Something like this:
```js
defaults = {
  ...
  submitCountryCode: false,
  submitCountryCodeName: 'country_code',
  ...
}
...
_generateMarkup: function() {
  ...
  if (this.options.submitCountryCode) {
    this.hiddenCountryInput = $("<input>", {
        "type": "hidden"
        "name": this.options.submitCountryCodeName
    }).appendTo(this.telInput);
  }
}
...
_selectListItem: function(listItem) {
  ...
  if (this.options.submitCountryCode) {
    this.hiddenCountryInput.val(listItem.attr("data-dial-code"));
  }
}
```

Let me know if this is a mergeable feature so I push a PR.

Thanks!

jackocnr:
Hi there, thank you for your suggestion and your pull request. First I need to better understand your use case. Why do you need to store both the national number and the country code on the backend?

If you look at the Troubleshooting section of the readme, I explain how to send the full international number to the backend, which is the recommended thing to do (then you only ever need to store/send 1 value). Then when you want to display that number to the user again, the plugin will take that full international number, and it will automatically select the right flag and convert the number to the national format to display to the user.

jackocnr:
Closing due to inactivity.

fedeisas:
@jackocnr the hidden input trick felt like a hack. This was just to provide an out of the box way of doing that.

I think the `geocomplete` plugin approach to this is really great: http://ubilabs.github.io/geocomplete/examples/form.html

In our system we store `country`, `number` (w/ area code) and `full_normalized_number`. I'm aware there are several approaches and best practices regarding how to store phone numbers in a DB.

[...]

PR #179, opened 2015-03-08
[#176] New submitCountryCode option

fedeisas:
# New submitCountryCode option

Issue: #176

New submitCountryCode option to keep a hidden input with the current dial code. Useful when using `nationalMode` but still want to submit the dropdown country dial code.

## Changelog

- Added to new options `submitCountryCode` and `submitCountryCodeName`.
- Hook into `_generateMarkup` method to create the hidden input with `submitCountryCodeName` option as input name.
- `_updateDialCode` method will now update the hidden input value as well.
- Added example and spec.

PS: I'm no good with jasmine, so you might want to add some more tests to it :smile:


fedeisas:
@jackocnr Travis is failling because I didn't commit the build version (it caused merge conflicts). Thanks.

jackocnr:
Hi there, thanks for coding this up, but I would like to discuss this in the issue before I consider including it. I'm going to close this for now, but if you reply to my questions in the issue, then we can come back to this. Thanks.

PR #180, opened 2015-03-09
Update intlTelInput.js

lucaspfsd:
Into _loadAutoCountry function, call function autoCountryLoaded by auto reference called "that", because is a internal call you don't need call as external function.


jackocnr:
Hey thanks for trying to help, but the idea with the current solution is that if you initialise the plugin on multiple inputs with the defaultCountry=auto option, then just the first one actually does the ipinfo call, and then when it's done it alerts all of them to the fetched country.

But also: don't keep working on this stuff as I intend to refactor it real soon!

PR #197, opened 2015-04-02
Set the default country based on input's initial value

### LINKED ISSUE #196: Zimbabwe flag is selected if initial value is cleared
If the input has an initial value, and you clear it (e.g. by selecting the text and pressing the delete key), the Zimbabwe flag is selected instead of it remaining on the current flag.

I've put together a test case here:  https://www.cloudmutt.com/intl-tel-input-bug/

I have a possible fix and will submit a pull request shortly.

AGWA:
This fixes #196.


jackocnr:
Thanks for putting this together, but I found a neater solution.

PR #201, opened 2015-04-08
Make preferredCountries work on mobile

nebirhos:
As title says, preferredCountries now appears on top of native select.


jackocnr:
Thanks for this. After doing a bit of research however, it seems that `<optgroup>` is supposed to be used to surround `<option>` elements, in order to create nested lists, which is not what we want. What about something like this: http://blog.jsfmp.com/post/97827050536/easy-attractive-dividers-html-select-elements-drop-downs

nebirhos:
Hi @jackocnr !
I used an empty \<optgroup\> because there's no clean way to add a divider on a select. Here is the result on Android:
empty

The "disabled line" trick is problematic on mobile because you cannot know the exact width of the screen, so the line remains smaller or even worse it breaks on two lines. On Android also it's displayed along with a faded radio button (see below).
line

Properly using \<optgroup\> looks really ugly on an Android device, all the text is indented:
optgroup

Maybe the cleanest way is to not put a divider at all:
none

jackocnr:
Thanks for doing this research! It appears there is no good solution. What about going back to a custom dropdown (same as for desktop), but we just make it full screen?

nebirhos:
Yep it would be great, but I don't have the time to develop such a feature from scratch.
The native menu with the right order it's sufficient for me now.

jackocnr:
Created a new issue for this: https://github.com/Bluefieldscom/intl-tel-input/issues/244

PR #203, opened 2015-04-09
Add "expanded" class to `div.intl-tel-input` when dropdown is opened

brunops:
- Use case is: need to change the border style of the input field when the flags are showing (set proper `border-radius` attributes), but the only class currently added is to `ul.country-list`


jackocnr:
Thanks for putting this together. Can you talk a bit more about why you want to change the border-radius when the dropdown is open? TBH it sounds like it might look a bit weird to keep changing the border-radius of an input when the user is interacting with it. And can you give an example of another populer lib that does something like this e.g. does Bootstrap do anything like this?

brunops:
Fair point, let be give you a better context..

Today, it seems like it's not possible to have different styles for the `input` element based on if the dropdown is expanded or not, because there's no class in a common parent.

The demo is like this:
screen shot 2015-04-09 at 12 54 07 pm

But if you increase the `border-radius` of the `input` field to more than `2px`, you'll see this 
screen shot 2015-04-09 at 12 54 29 pm

With a common parent class, it would be possible to style it and have a final layout like 
screen shot 2015-04-09 at 12 54 45 pm

I'm having to do this for a current project, so I thought about giving it back. Btw, already noticed an issue with my PR, because the dropdown might appear down or above the class needs to reflect the position too. Let me know if you think this is a good idea or not so I can close or update the PR.

jackocnr:
Thanks a lot for the idea and the code, but no-one has requested this feature before, so I'm reluctant to merge it at this point. TBH I'm not sold on the use-case (unless you can provide examples of other respected people/plugins doing something similar?), so I think I'll close this for now, and see if there's more demand for it in the future.

Issue #195, opened 2015-04-01
Country code select tab navigation


brunops:
First of all, thanks for this amazing plugin. 
Sorry if this is a duplicate, but couldn't find this topic in the README or the closed issues. I'm currently having one problem when tab navigating the form, which is the flags dropdown not getting focus (probably because it is a `ul` element, and not a form field), is there any way around it?

Thanks

jackocnr:
Yes I agree: ideally you would be able to tab to the dropdown, but yes you're right: it would only work if the dropdown was a normal `<select>` element, which we can't do, as we'd lose all the nice styling. I know some browsers support custom styling of `<select>` elements, but I know it's difficult (different in each browser), and limited (not sure if we could have flags in the dropdown, not sure if we could style the selected flag in the input to look right, and I know you can't do it at all in older versions of IE). Needs more research.

kirkas:
I suggest adding the `tabindex` attribute on the `.selected-flag`, ditch the `click`event and use the `focus` & `blur` events and accordingly show/hide the country-list.

I manage to make this work with a quick hacky play around on the demo page:

![screen shot 2015-04-01 at 11 09 59 am](https://cloud.githubusercontent.com/assets/1159783/6948022/e700f4e6-d85f-11e4-92ed-1408cf47c922.png)

I think this should be the default plugin behavior as this is a huge improvement for accessibility (tab & click VS click only).


Ref for making an element tabbable:
http://stackoverflow.com/questions/13637223/how-do-you-make-a-div-tabbable

kirkas:
Here's what it look like if you put `tabindex` and `focus/blur` event on the `.flag-dropdown` instead of the `.selected-flag`

![screen shot 2015-04-01 at 11 25 07 am](https://cloud.githubusercontent.com/assets/1159783/6948299/c8ea8bb4-d861-11e4-9d4b-888679695956.png)

Theoretically, you could even tab trough the list of country, or just the preferred country as tabbing trough 235 element might be quite painful for the user.

brunops:
Hey, thanks for the quick replies.

@jackocnr 
Agreed, using a `<select>` element is probably not an option.

@kirkas 
You're right! This might be the way around this, I personally would expect it to behave just like a `<select>` element, so `tabindex` would only be in `.selected-flag` or `.flag-dropdown`, I'll try to play around with it.

Thanks.

jackocnr:
[...]

PR #213, opened 2015-04-21
Accessibility

brunops:
Make plugin accessible for tab navigation, using idea discussed in #195
Tried to follow all the code styles, added comments and tests.

The commit messages have detailed information of each change.
Let me know your thoughts.


jackocnr:
Wow this is amazing thank you.

I just had a quick look and currently on Chrome, when you tab to the selected country, the blue border seems to go under the `input` on the right. There are 2 solutions to this (off the top of my head): either just add `z-index: 1;` to the `.flag-dropdown` (we need to consider this carefully as we've had z-index issues in the past with multiple instances overlapping each other), OR move the `tabindex` thing from the `.flag-dropdown` to the `.selected-flag`, which changes the blue outline thing when the dropdown is open: currently the whole dropdown has the blue outline, but if you move the tabindex, then it's only ever the selected-flag that has it. I'm not sure which makes more sense. Thoughts?

brunops:
Cool! I'm glad you liked it.
Good catch, IMHO it makes more sense to keep the `tabindex` in `.flag-dropdown`, as you've mentioned it keep the blue outline over both the select and the list, which looks nice.
Adding the `z-index: 1;` to `.flag-dropdown` seems to have fixed it, but I'm not aware of the issue you've just mentioned, would you remember the steps to reproduce it?

jackocnr:
Just have 2 inputs: one above the other, and instantiate the plugin on both, and check that opening the dropdown on the top goes over the bottom one properly. Does that make sense? The dropdown of the first input should go over the selected flag of the second input

brunops:
Yes @jackocnr, I understand what you're saying now..
screen shot 2015-04-23 at 9 40 28 am

Not sure how to solve it.. only CSS fix I could think of was to have incremental `z-index` properties for each of the instances on the page (so first would be `1`, second `2` and so on..), but I don't like this solution, it'll probably introduce other issues.

Should we go with the `tabindex` on `.selected-flag` approach instead?

jackocnr:
Yer I've looked into it and I don't think there's a solution to the z-index problem, so yer let's try moving the tabindex and I'll test that solution some more. Thanks man.

brunops:
Just made the update, the outline is not as great as before, but the accessibility part is probably a bigger win.. let me know your thoughts.

jackocnr:
Thank you so much!

PR #216, opened 2015-04-27
Fix comments

brunops:
Just noticed that the comments didn't move with the `tabindex` element change, sorry about that..


jackocnr:
Thanks

PR #210, opened 2015-04-20
adds support to multiple geoip data providers

zvictor:
ipinfo is too limited, and options are nice :)

PR #218, opened 2015-05-04
adds support to multiple geoip data providers

ruhley:
Based off #210 by @zvictor to allow people to choose different geoip providers. I also added the possibility to add a custom function to use for any geoip provider.


adisos:
Please, merge this. :)

jackocnr:
Sorry I've taken so long to respond. I don't like the idea of having a list of "supported" providers, which is arbitrary, prone to breaking, and very limited in options e.g. in the original code, we do a jsonp request to ipinfo.io, but there's no option to do that the new way. I do however like the idea of having a custom function to handle this mess outside of the plugin - in fact I think this is much better than the original solution.

So if you update this pull request to just be about the custom function, then I'll merge it in! Thanks.

Tim-Intellipharm:
I have removed the default geoip providers and left only a function if the user provided it. The only thing I can think that you may wish to change is that the option is called ```geoipProvider```. I couldn't think of anything else to name it though.

jackocnr:
Thank you for this! Let's call the option `geoIpLookup` for now.

Also:
- Let's just have 1 example in the readme, as that file is already so long. We can add more in the wiki if people need.
- Please use `that` instead of `self` for consistency
- Please inline the `_autoCountryLoaded` function as it's only called from that one place
- Instead of `resp` or `country_code` let's call the argument `countryCode` for consistency

ruhley:
I have updated the code to your requests.

build/js/intlTelInput.js
```diff
@@ -358,23 +358,26 @@ https://github.com/Bluefieldscom/intl-tel-input.git
...
+            if ($.cookie) {
+                $.cookie("itiAutoCountry", $.fn[pluginName].autoCountry, {
+                    path: "/"
                 });
             }
+            // tell all instances the auto country is ready
+            // TODO: this should just be the current instances
+            // $(".intl-tel-input input").intlTelInput("autoCountryLoaded");
+            // Why not just call it directly??
+            this.autoCountryLoaded();
```
jackocnr: We cannot just call `autoCountryLoaded` directly, because if there were any other inputs instantiated at the same time as this one, they will also be waiting for this callback (e.g. if you instantiated the plugin on multiple inputs at once - see line 419 where we check `startedLoadingAutoCountry` - if this is set to true, then those instances don't do anything - they are just waiting). Does that make sense?

ruhley: Yea, that makes sense. I thought there would be some reason you weren't doing this to begin with. I have updated the code to remove that.

jackocnr:
Sorry that took so long! Thanks again for the contribution.

PR #222, opened 2015-05-13
add intl-tel-mobile-select class for mobile select elements

tripflex:
I have found multiple themes that modify or change select elements and without having a class it makes it difficult to override or have theme developers add to exclude list.  I noticed the `<ul>` includes a class but the `<select>` does not, so added one to help with these issues.

Add `intl-tel-mobile-select` class to select element for mobile devices

### DIFF
--- a/src/js/intlTelInput.js
+++ b/src/js/intlTelInput.js
@@ -223,7 +223,7 @@ Plugin.prototype = {
     // desktop is a proper list containing: preferred countries, then divider, then all countries
     if (this.isMobile) {
       this.countryList = $("<select>", { 
-        "class": "intl-tel-mobile-select"
+        "class": "iti-mobile-select"
       }).appendTo(this.flagsContainer);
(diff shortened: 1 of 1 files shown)


jackocnr:
Hmm why can't you just use the selector: `.intl-tel-input select {...}`?

Can you give me a specific example of a theme and the exact problem you're having?

tripflex:
The specific theme i'm referring to would be Listify and the reason is because they convert all select boxes to chosen and in order to have them add to the array of classes to skip I needed a class for the select item.  Just figured since the `<ul>` for the list has a class, the `<select>` for mobile should have one too

jackocnr:
Ok. If you update the class to `iti-mobile-select` to be more consistent with new class names I want to use then I will merge it in. Thanks!

tripflex:
Thanks man, updated to new class name :+1:

Issue #237, opened 2015-06-15
[Proposal] Make Sass variables overridable


justin-lau:
It would help developers to customize the style of this plugin by making all the Sass variables overridable. Especially the image paths and filenames, so that when a consuming application uses Sass they just need to override the variables.

It may look something like this:

```scss
// add `!default` to existing variables to allow overriding
$hoverColor: rgba(0, 0, 0, 0.05) !default;

// ...

// add image related variables for even more convenience
$flagsImagePath: "../img/" !default;
$flagsImageName: "flags" !default;
$flagsImageExtension: "png" !default;

// ...

.iti-flag {
  // ...

  background-image: url("#{$flagsImagePath}#{$flagsImageName}.#{$flagsImageExtension}");
 
  // ...

  @media only screen and (-webkit-min-device-pixel-ratio: 2), only screen and (min--moz-device-pixel-ratio: 2), only screen and (-o-min-device-pixel-ratio: 2 / 1), only screen and (min-device-pixel-ratio: 2), only screen and (min-resolution: 192dpi), only screen and (min-resolution: 2dppx) {
    background-image: url("#{$flagsImagePath}#{$flagsImageName}@2x.#{$flagsImageExtension}");
  }

}

```

When an application uses this plugin, they can actually import the `intlTelInput.scss` file and override the variables to customize the styling:

```scss
// overriding $hoverColor
$hoverColor: rgba(red, 0.05);

// overriding image path
$flagsImagePath: "assets/images/";

@import "bower_component/intl-tel-input/src/css/intlTelInput";
```

PR #238, opened 2015-06-15
Issue #237: Made variables overridable

justin-lau:
Please refer to #237 for details of this PR.

It only makes the existing variables overridable and adds a few variables for the image path and filename. 

Refactored the retina image url part a little to make better use of Sass's nesting syntax, but it doesn't affect the compiled css at all.

### DIFF
--- a/src/css/intlTelInput.scss
+++ b/src/css/intlTelInput.scss
@@ -1,3 +1,18 @@
+/**
+ * Variables declared here can be overridden by consuming applications, with
+ * the help of the `!default` flag.
+ *
+ * @example
+ *     // overriding $hoverColor
+ *     $hoverColor: rgba(red, 0.05);
+ *
+ *     // overriding image path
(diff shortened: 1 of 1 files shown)


jackocnr:
This looks great thanks. If you could just add a comment at the top about why we're using `!default` so I don't forget, then I will merge this in!

src/css/intlTelInput.scss
```diff
@@ -1,25 +1,44 @@
...
+ *     // overriding $hoverColor
+ *     $hoverColor: rgba(red, 0.05);
+ *
+ *     // overriding image path
+ *     $flagsImagePath: "images/";
+ *
+ *     // import the scss file after the overrides
+ *     @import "bower_component/intl-tel-input/src/css/intlTelInput";
+ */
+
```
justin-lau: How does this sound?

Issue #246, opened 2015-06-23
Display the country list on top of everything


seroche:
Hi,

First I want to thank you for the great quality of your work. Your lib is easy to use and very user friendly. 

However, i have one small problem:

![image](https://cloud.githubusercontent.com/assets/12636347/8297920/8f10b73c-1994-11e5-9109-b4263380b4b5.png)

As you can see the country list is displayed within my div and not on top of everything. 
> Is there a way to do that so the user can see the full list without having to scroll down?

Browser: IE11 (IE8 compatibility mode due to Sharepoint 2010....)

Thank you,
Seb

caseyjhol:
This would probably need to be resolved by adding a `container` option and setting it to `'body'` similar to Bootstrap popover (http://getbootstrap.com/javascript/#popovers-options). I implemented the same feature in Bootstrap-Select and am in the process of working on a pull request for this right now.

jackocnr:
Hi guys. @caseyjhol I'm interested to see what you come up with in your pull request, but as far as I can see this is just a simple CSS issue relating to `z-index` and `overflow` and is outside of the scope of the plugin, because it depends on how you use these CSS rules on the other elements on the page.

caseyjhol:
If you have the phone input inside a div with `overflow: scroll` or `overflow: hidden`, currently it is impossible for the intl-tel-input dropdown to ever appear outside the div. See http://silviomoreto.github.io/bootstrap-select/#container for an example of the problem/fix.

jackocnr:
Hmm interesting. Your code may also be useful for this mobile dropdown issue: https://github.com/Bluefieldscom/intl-tel-input/issues/244 where we plan to move back to a custom dropdown, but would need it to be full-screen, which would mean injecting it at the end of the body.

caseyjhol:
Definitely - I'm still familiarizing myself with your code, but I should be able to get a pull request/working demo within a few days.

caseyjhol:
I'll need to clean up the code a bit before the pull request, but check out: http://embed.plnkr.co/lobCJwstCUnRHx5FxZ9s/preview.

A total of about 33 new lines of JS and a few new lines of CSS. This could be easily adapted to a "full screen" implementation to resolve #244.

jackocnr:
Excellent work! Thank you. I look forward to seeing the code.

caseyjhol:
[...]

PR #247, opened 2015-06-26
Add dropdownContainer option to append country list to a specific element (#246)

caseyjhol:


caseyjhol:
fixed merge conflicts

jackocnr:
Solves https://github.com/Bluefieldscom/intl-tel-input/issues/246

src/css/intlTelInput.scss
```diff
@@ -229,6 +229,11 @@ $flagsImageExtension: "png" !default;
     height: 100%;
     opacity: 0;
   }
+
+  // if container option is set, increase z-index to prevent display issues
+  &.iti-container {
+    z-index: 1060;
```
jackocnr: What made you choose this z-index value?

jackocnr:
Thank you very much for coding this up.

I'm not sure about the need for allowing any custom container for the dropdown - instead why don't we just have an option to put the dropdown at the end of the body? I think that keeps things simpler for the dev and simpler for the plugin logic (because if the container has position: relative, wont that break the positioning of the dropdown?).

src/js/intlTelInput.js.ejs
```diff
@@ -761,6 +765,33 @@ Plugin.prototype = {
...
+        "left": pos.left,
+        "position": "absolute"
+      });
+      
+      // close menu on resize/scroll
+      this.telInput.parents().off("resize" + this.ns + " scroll" + this.ns).on("resize" + this.ns + " scroll" + this.ns, function () {
+        if (!that.countryList.hasClass("hide")) that._closeDropdown();
+      });
+      
+      this.$drop.appendTo(this.options.container);
```
jackocnr: I see you inject the markup into the container in `_setDropdownPosition`, which is called every time you open the dropdown. Instead couldn't we just do this once, in `_generateMarkup`, and just show/hide it each time (and update the CSS top/left values)?

src/css/intlTelInput.scss
caseyjhol: The default bootstrap modal z-index is 1050. 1060 allows the menu to appear above a modal and should cover most use cases and can always be customized if need be. A z-index should be set by default, though, to make this feature usable out of the box, but I'm not opposed to changing it.

src/js/intlTelInput.js.ejs
caseyjhol: I'm injecting the menu on every open and using `.detach()` on every close (same as remove except all jQuery data is kept associated with the menu when it is reinserted). This prevents the DOM from being polluted and works better on single page apps. Regenerating it on every open is only about a 1-2 ms difference, but I should move some of that code to _generateMarkup to completely remove that lag.

Also, since we're already using the `iti-container` class, it would make more sense to just set `position: absolute` in the CSS. I'll update my code.

caseyjhol:
I think giving the dev the option to choose the container is the preferred method. It's the same amount of code on our end and gives the dev more flexibility. Also, that's how Bootstrap does it, and I think would be the expected implementation.

build/js/intlTelInput.js
```diff
@@ -191,6 +193,14 @@ https://github.com/Bluefieldscom/intl-tel-input.git
             $("<div>", {
                 "class": "arrow"
             }).appendTo(selectedFlag);
+            if (this.options.container) {
+                this.$drop = $("<div>", {
+                    "class": "intl-tel-input iti-container"
+                });
+                $("<div>", {
+                    "class": "flag-dropdown"
+                }).appendTo(this.$drop).append(this.countryList);
```
jackocnr: this.countryList is not defined yet. And also if you're going to append it to this div here, then you shouldn't append it to the flagsContainer later as well...

build/js/intlTelInput.js
```diff
@@ -631,11 +641,25 @@ https://github.com/Bluefieldscom/intl-tel-input.git
...
+            // if container calculate postion and append this.$drop to container
+            if (this.options.container) {
+                var inputHeight = !dropdownFitsBelow && dropdownFitsAbove ? 0 : that.telInput.outerHeight();
+                // calculate placement
+                this.$drop.css({
+                    top: pos.top + inputHeight - parseInt(this.telInput.css("borderTopWidth"), 10) - parseInt(this.telInput.css("borderBottomWidth"), 10),
+                    left: pos.left
+                });
+                // close menu on resize/scroll
+                this.telInput.parents().off("resize" + this.ns + " scroll" + this.ns).on("resize" + this.ns + " scroll" + this.ns, function() {
```
jackocnr: Is this specific to this new container feature? Unless this is specifically needed for this new feature, I think you should open a separate issue for this, and we should think about how to include it even if the dev doesn't use the container option.

build/js/intlTelInput.js
```diff
@@ -24,6 +24,8 @@ https://github.com/Bluefieldscom/intl-tel-input.git
         autoHideDialCode: true,
         // add or remove input placeholder with an example number for the selected country
         autoPlaceholder: true,
+        // append menu to a specific element
+        container: false,
```
jackocnr: This name is a bit generic - shall we call it dropdownContainer?

build/js/intlTelInput.js
caseyjhol: Are you referring to the resize/scroll event handler?

jackocnr: Yes

caseyjhol: When the window is scrolled/resized, the menu will be in the wrong position unless the position is constantly recalculated on scroll/resize. This is doable, but more resource intensive and also can create some weird issues when the input is scrolled out of frame. The menu would still be visible even though the input isn't. Play around with scrolling the inputs with `container: 'body'`.

The underlying code in these examples is outdated, but the concepts are the same:

[When scrolling, menu position is not recalculated](http://embed.plnkr.co/lobCJwstCUnRHx5FxZ9s/preview)
[When scrolling, menu position is recalculated](http://embed.plnkr.co/vYqhcKJGKxWp6vUvLTHZ/preview)

I figured it would best to simply close the menu. I suppose this would be something we could implement as an option (`autoClose`?) for devs not using the `dropdownContainer` option, but I'm not sure how desirable it would be to use `autoClose` without also using `dropdownContainer`, and likewise, how desirable it would be to use `dropdownContainer` without `autoClose`.

I guess we could just make `autoClose` enabled by default if a `dropdownContainer` is set, unless the dev specifies that it should be disabled to allow for full flexibility.

caseyjhol: I can remove this feature for now if you'd prefer.

caseyjhol:
Removed the event handler that automatically closes the menu on resize/scroll.

caseyjhol:
resolve merge conflicts

caseyjhol:
@jackocnr Anything you need in particular to get this merged?

jackocnr:
Hi I'm really sorry for the delay, I will add some comments now.

build/js/intlTelInput.js
```diff
@@ -216,6 +218,15 @@ https://github.com/Bluefieldscom/intl-tel-input.git
                 this.countryList.removeClass("v-hide").addClass("hide");
                 // this is useful in lots of places
                 this.countryListItems = this.countryList.children(".country");
+                // create dropdownContainer markup
+                if (this.options.dropdownContainer) {
+                    this.$drop = $("<div>", {
```
jackocnr: We don't use $ variable names anywhere else - please remove the $. Also let's call this variable "dropdown" instead of just "drop".

src/css/intlTelInput.scss
```diff
@@ -229,6 +229,12 @@ $flagsImageExtension: "png" !default;
     height: 100%;
     opacity: 0;
   }
+
+  // if container option is set, increase z-index to prevent display issues
+  &.iti-container {
+    position: absolute;
+    z-index: 1060;
```
jackocnr: Ok I'm down with copying Bootstrap's z-index value for now (for lack of any other ideas), but let's at least add a comment explaining why we went with this particular value for future reference.

jackocnr:
Regarding the resize/scroll event listener - I understand the need for it now. If you set `dropdownContainer` to body, but the input is in another scrollable element, then scrolling that element will move the input but not the dropdown. (similar issue for resize event).

So we cannot add this feature without providing some solution to this. But it seems like too much to add 2 event listeners on every ancestor of the input. What do Bootstrap do for tooltips?

build/js/intlTelInput.js
```diff
@@ -634,11 +645,21 @@ https://github.com/Bluefieldscom/intl-tel-input.git
...
             dropdownFitsBelow = inputTop + this.telInput.outerHeight() + this.dropdownHeight < windowTop + $(window).height(), dropdownFitsAbove = inputTop - this.dropdownHeight > windowTop;
             // dropdownHeight - 1 for border
             var cssTop = !dropdownFitsBelow && dropdownFitsAbove ? "-" + (this.dropdownHeight - 1) + "px" : "";
             this.countryList.css("top", cssTop);
+            // if container calculate postion and append this.$drop to container
+            if (this.options.dropdownContainer && !this.isMobile) {
+                var inputHeight = !dropdownFitsBelow && dropdownFitsAbove ? 0 : this.telInput.outerHeight();
+                // calculate placement
+                this.$drop.css({
+                    top: pos.top + inputHeight - parseInt(this.telInput.css("borderTopWidth"), 10) - parseInt(this.telInput.css("borderBottomWidth"), 10),
```
jackocnr: If you want the height without the borders, then just use innerHeight instead of outerHeight

jackocnr:
Also one other thing I noticed: if you just run the included demo.html, except you give the form element margin-top:1000px; so that the dropdown appears above the input, then you can see a 1px gap between the input and the dropdown.

caseyjhol:
We could do something like:

```js
// ...
defaults = {
// ...
scrollListener: 'body',
// ...
}
// ...
var scrollListener = this.options.scrollListener === 'body' ? $('body') : this.telInput.closest(this.options.scrollListener);

scrollListener.off("scroll" + this.ns).on("scroll" + this.ns, function () {
  if (!that.countryList.hasClass("hide")) that._closeDropdown();
});
```

caseyjhol:
Also, regarding the gap that is created when adding `margin-top:1000px;` to the form element, that was fixed after changing to `innerHeight()` instead of `outerHeight()` and subtracting borders. The borders were *always* being subtracted, even when inputHeight was 0, so the menu was 2px higher than it should have been.

jackocnr:
Thanks for the updates!

Re: scrolling, it works fine if dropdownContainer is set to the body, and you scroll the body. Would it work if we just said that if they put the input inside a scrollable element, and they have to set that element as the dropdownContainer element? Would that solve the problem?

Do you know what Bootstrap do?

caseyjhol:
I looked at Bootstrap's code, and it looks like right now, they don't do anything about it. They leave refreshing the tooltip up to the user. Setting the scrollable parent element as the dropdownContainer wouldn't work unfortunately, as the dropdown needs to be in an element outside of the scrollable element to remain visible (otherwise the overflow of the scrollable element would cut it off, leaving us with our original issue).

Bootstrap 4 (currently in alpha) fixes this by utilizing a framework called [Tether](https://github.com/HubSpot/tether).

In bootstrap-select, the event listener is only on 'body', and then in my own code I do

```js
$('#scrollingElement').scroll(function() {
    $('body').trigger('scroll');
});
```
My code above https://github.com/Bluefieldscom/intl-tel-input/pull/247#issuecomment-131899433 somewhat streamlines this by bypassing the scroll listener on the `body` element only, and allowing the user to add the event listener directly on the element they're scrolling. It does seem a bit like overkill, though, to create another option just for the dropdownContainer. Perhaps we could do an array or object? For example, if you set `dropdownContainer: ['body', '#scrollingElement']`, then it appends the menu to `body` and listens for the scroll on `#scrollingElement`, but if it's just a string `dropdownContainer: 'body'` then it appends to `body` and listens for the scroll on `body`.

caseyjhol:
Any thoughts?

jackocnr:
I say we just listen to the scroll event on the body (and close the dropdown), and leave the user to add their own scroll listeners themselves if they want to do anything more complicated. 

I think that this will already be a useful feature as it is and we should get it out and get some more feedback and then see what people want before we worry about adding extra scroll watching functionality.

caseyjhol:
Works for me. I'll resolve the conflicts ASAP so we can get this merged.

caseyjhol:
Should be good to go

build/js/intlTelInput.js
```diff
@@ -636,11 +647,25 @@ https://github.com/Bluefieldscom/intl-tel-input.git
...
+            // if container calculate postion and append this.dropdown to container
+            if (this.options.dropdownContainer && !this.isMobile) {
+                var inputHeight = !dropdownFitsBelow && dropdownFitsAbove ? 0 : this.telInput.innerHeight();
+                // calculate placement
+                this.dropdown.css({
+                    top: pos.top + inputHeight,
+                    left: pos.left
+                });
+                // close menu on body scroll
+                $("body").off("scroll" + this.ns).on("scroll" + this.ns, function() {
```
jackocnr: Shouldn't the $("body").off("scroll") go in _closeDropdown instead?

And then whenever the on scroll event listener fires, you know the dropdown is open, so you don't need to check if countryList.hasClass("hide").

build/js/intlTelInput.js
```diff
@@ -218,6 +220,15 @@ https://github.com/Bluefieldscom/intl-tel-input.git
...
                 // this is useful in lots of places
                 this.countryListItems = this.countryList.children(".country");
+                // create dropdownContainer markup
+                if (this.options.dropdownContainer) {
+                    this.dropdown = $("<div>", {
+                        "class": "intl-tel-input iti-container"
+                    });
+                    $("<div>", {
+                        "class": "flag-dropdown"
+                    }).appendTo(this.dropdown).append(this.countryList);
```
jackocnr: We've already appended this.countryList to this.flagsContainer earlier in this function - if we're appending it here then we shouldn't do that. We should put a condition around that earlier line so we don't do it twice! (see the line where we create this.countryList)

jackocnr:
Thanks for your great work! Any chance you could update your example, so I can play with it?

caseyjhol:
http://plnkr.co/edit/V4qUWx8X4BKhh56cbaBo?p=preview

build/js/intlTelInput.js
caseyjhol: this.countryList needs to be appended to a visible element at some point prior to opening the menu in order to properly determine this.dropdownHeight. It'll take some code rearrangement to not append this.countryList to this.flagsContainer when the markup is generated. Appending countryList will have to be moved to _setDropdownPosition.

The nice thing is, it doesn't actually add any additional load time, but I do agree we could probably optimize it a bit further. Although, appending it twice only adds about 20ms per input, so we might be in the realm of micro-optimization.

caseyjhol: Working on a new method right now - improves load time for inputs both with and without dropdownContainer.

caseyjhol: See it in action: http://plnkr.co/edit/V4qUWx8X4BKhh56cbaBo?p=preview

build/js/intlTelInput.js
```diff
@@ -636,11 +644,27 @@ https://github.com/Bluefieldscom/intl-tel-input.git
         },
         // decide where to position dropdown (depends on position within viewport, and scroll)
         _setDropdownPosition: function() {
-            var inputTop = this.telInput.offset().top, windowTop = $(window).scrollTop(), // dropdownFitsBelow = (dropdownBottom < windowBottom)
-            dropdownFitsBelow = inputTop + this.telInput.outerHeight() + this.dropdownHeight < windowTop + $(window).height(), dropdownFitsAbove = inputTop - this.dropdownHeight > windowTop;
+            var showDropdownContainer = this.options.dropdownContainer && !this.isMobile;
+            if (showDropdownContainer) this.dropdown.appendTo(this.options.dropdownContainer);
+            var that = this, pos = this.telInput.offset(), // show the menu and grab the dropdown height
+            dropdownHeight = this.countryList.removeClass("hide").outerHeight(), inputTop = pos.top, windowTop = $(window).scrollTop(), // dropdownFitsBelow = (dropdownBottom < windowBottom)
```
jackocnr: Hmm but if there's no dropdownContainer, then countryList is already in the DOM and if you do removeClass("hide") then it will be visible! You need to addClass("v-hide") first to get the height. Also we only need to do this once, so I recommend checking if (this.dropdownHeight) is set, and if not, doing addClass("v-hide").removeClass("hide") and setting this.dropdownHeight and then doing addClass("hide").removeClass("v-hide") again. This will only need to be done once... What do you think?

jackocnr:
Thanks for the Plunker! Is it supposed to close the dropdown on scroll? Because that's not working for me.

Update: oh it might be because you're using float for positioning, and body has height:0?

build/js/intlTelInput.js
caseyjhol: With this method, it's not necessary to add/remove the `v-hide` class. I moved `this.countryList.removeClass("hide");` from the `_showDropdown` function into the `_setDropdownPosition` function. This allows us to calculate dropdownHeight before doing anything else in `_setDropdownPosition` (but after the dropdownContainer has been appended). So, yes, the menu will be visible, but that's what we want, as it allows us to get this.countryList's outerHeight.

Basically, instead of adding `v-hide` and removing `hide`, then calculating the height, and then removing `v-hide` and adding `hide`, and then later removing `hide` when the menu is open, we're just calculating the height when `hide` is removed, as we have to do that anyway.

I thought about caching dropdownHeight after it's calculated (so it doesn't need to be recalculated on subsequent openings), but we have to do `removeClass("hide")` anyway, and doing `outerHeight()` isn't really a performance hit. This also leaves open the option in the future to allow for more dynamic menu heights - cases where it might be desirable to have the menu height change between openings (like if countries have been removed from or added to the list).

caseyjhol:
Hrmm - make sure you're checking the right plunker demo - http://plnkr.co/edit/V4qUWx8X4BKhh56cbaBo?p=preview.

Scrolling within either of the gray boxes with `container: 'body'` set should close the menu (or at least it works for me in Windows 10 and Chrome/Firefox.

seroche:
Hi Guys,

Stupid question: how is it supposed to work?

I tried to add an option:
dropdownContainer: 'body'
or
dropdownContainer: document.body
> Not working in IE8 or Chrome. Am I missing something?

Thank you,
Sebastien

caseyjhol:
Seroche - can you provide me an example of it not working? You're sure you're using my code, and not the current intl-tel-input code?

jackocnr:
@caseyjhol you say scrolling within the grey boxes causes the dropdown to close, but I don't understand: I thought the scroll listener was on the body and should react to the body scrolling, not the grey boxes?

caseyjhol:
@jackocnr In the intl-tel-input code, we are listening for the scroll on the body. We decided this was the best method for now - https://github.com/Bluefieldscom/intl-tel-input/pull/247#issuecomment-133097668. In the script.js file in Plunker, I'm running

```js
  $('.input-container').on('scroll', function() {
    $('body').scroll();
  });
```

`input-container` is a class on the gray boxes. That means that scrolling in the gray box triggers a scroll on the body, and then closes the dropdown. In any case, `'body'` should be switched to `window`, as that seems to be the most cross-browser compliant.

caseyjhol:
The Plunker has been updated to listen for the scroll on `window` instead of `'body'`.

caseyjhol:
Conflicts resolved.

jackocnr:
Hi @caseyjhol, thanks again for your work on this - I think we're nearly there. The window scroll thing is now working, which is great.

That said, I'm not happy with the changes to `_setDropdownPosition`. In your plunker, if you set a breakpoint on line 653 of intlTelInput.js and then set the window size/scroll so that the dropdown should open above the input, then click to open the dropdown: when it pauses at the breakpoint, you will see that at that point we are already showing the country list wherever it is currently positioned (if there is no dropdownContainer set, then this is directly below the input like normal, or if there is a dropdownContainer set then you can see it at the bottom of the page). This is before we do the calculations and move it to above the input.

I propose we go back to the original (existing) solution of appending the countryList in the regular place (next to the input) during initialisation and caching the height, and then if dropdownContainer is set, then we detach it.

I also think the `flag-dropdown` duplication is messy and unnecessary. Let's remove it from the dropdownContainer markup (I don't think it is needed), and let's also rename the original one (around the `selected-flag` element) to be called `flag-container` instead - I think that makes more sense.

Finally, this positioning stuff is getting complicated - we need some comments! Please add this first comment before the `cssTop` line: 

    // by default the dropdown will be below the input (due to the height of the selected flag element). If we want to position it above the input, we need to change it's "top" value. Setting top:0 would position it directly over the input, so we set it to minus it's height so it appears above.

And this comment before the `inputHeight` line (please rename that variable to `extraTop`):

    // by default the dropdown will be directly over the input because it's not in the flow. If we want to position it below, we need to add some extra top value.

Then I think we might be ready to merge it in! Thanks again!

caseyjhol:
>In your plunker, if you set a breakpoint on line 653 of intlTelInput.js and then set the window size/scroll so that the dropdown should open above the input, then click to open the dropdown: when it pauses at the breakpoint, you will see that at that point we are already showing the country list wherever it is currently positioned (if there is no dropdownContainer set, then this is directly below the input like normal, or if there is a dropdownContainer set then you can see it at the bottom of the page). This is before we do the calculations and move it to above the input.

1. Putting a breakpoint there shows the menu as being visible, and in the incorrect location, but this will never happen. The `hide` class is removed from `this.countryList`, `dropdownHeight` is calculated, the `that, pos, inputTop, windowTop, dropdownFitsBelow, dropdownFitsAbove,` and `cssTop` variables are set, and then `this.countryList.css("top", cssTop)` is run. All of this happens simultaneously, meaning that the user never sees the unstyled menu - the browser never even renders it in the incorrect position because it happens so fast.

    I've tested caching dropdownHeight in Chrome, Firefox, Internet Explorer 11, and Microsoft Edge, and in every case, caching made no performance improvement. We need to remove the `hide` class from `this.countryList` every time anyway, so retrieving the height via `outerHeight` adds no additional load time, and thus it is unnecessary to cache dropdownHeight. I've reformatted the code a bit more (I'll update soon), and opening the menu (both with and without dropdownContainer) is now as fast as opening the menu is in the current live code in the master branch.

2. The dropdownContainer markup includes the `flag-dropdown` div to keep styling the same.
>from intlTelInput.scss#L98

    >.intl-tel-input .flag-dropdown {
        position: absolute;
        top: 0;
        bottom: 0;
        padding: $borderWidth;
}

    Although, I can move `padding: $borderWidth;` to `.iti-container` and then we could skip the duplication.

jackocnr:
> All of this happens simultaneously, meaning that the user never sees the unstyled menu - the browser never even renders it in the incorrect position because it happens so fast.

Are you 100% sure that the dropdown will never render in the wrong place? Even on an old, slow computer? How can you know? I don't know much about browser rendering, so I'm actually asking. If you can convince me that there's no way the user will ever see a flash of incorrect positioning, then I'm happy to go ahead with your solution. If there's a chance some people might see it, then I think it's worth doing the calculations up front, before we unhide it.

caseyjhol:
A quick way to check is to use Timeline in Chrome DevTools. See https://developers.google.com/web/fundamentals/performance/rendering/simplify-paint-complexity-and-reduce-paint-areas?hl=en. If you enable "Show paint rectangles", you'll see that the menu is never painted in the upper left corner of the screen, and there's never a "Recalculate style", meaning the menu is only ever positioned in the correct spot on the screen.

I worded my sentence poorly. It's not that this happens so fast that the browser doesn't render it, it's that the browser never has a reason to render it. I'll try to explain. I've included the `_setDropdownPosition` function with line numbers below for easy reference.

```js
1   // decide where to position dropdown (depends on position within viewport, and scroll)
2   _setDropdownPosition: function() {
3     var showDropdownContainer = this.options.dropdownContainer && !this.isMobile;
4 
5     if (showDropdownContainer) this.dropdown.appendTo(this.options.dropdownContainer);
6 
7     // show the menu and grab the dropdown height
8     this.dropdownHeight = this.countryList.removeClass("hide").outerHeight();
9 
10    var that = this,
11      pos = this.telInput.offset(),
12      inputTop = pos.top,
13      windowTop = $(window).scrollTop(),
14      // dropdownFitsBelow = (dropdownBottom < windowBottom)
15      dropdownFitsBelow = (inputTop + this.telInput.outerHeight() + this.dropdownHeight < windowTop + $(window).height()),
16      dropdownFitsAbove = (inputTop - this.dropdownHeight > windowTop),
17      // by default, the dropdown will be below the input (due to the height of the selected flag element). If we want to position it above the input, we need to change it's "top" value. Setting top: 0 would position it directly over the input, so we set it to minus it's height so it appears above.
18      cssTop = (!dropdownFitsBelow && dropdownFitsAbove) ? "-" + (this.dropdownHeight - 1) + "px" : "";
19
20    this.countryList.css("top", cssTop);
21    
22    // if dropdownContainer is enabled, calculate postion
23    if (showDropdownContainer) {
24      // by default the dropdown will be directly over the input because it's not in the flow. If we want to position it below, we need to add some extra top value.
25      var extraTop = !dropdownFitsBelow && dropdownFitsAbove ? 0 : this.telInput.innerHeight();
26      
27      // calculate placement
28      this.dropdown.css({
29        "top": inputTop + extraTop,
30        "left": pos.left
31      });
32
33      // close menu on window scroll
34      $(window).on("scroll" + this.ns, function () {
35        that._closeDropdown();
36      });
37    }
38  },
```
Basically, the `_setDropdownPosition` function does 2 things - it allows `this.countryList` to be visible, and it puts it in the correct position on the page. None of the code in lines 10-18 affects layout at its time of execution as far as the browser is concerned. So, while those variables are used later on to correctly position `this.countryList`, they are simply calculations, unlike, for example, the `this.dropdownHeight` declaration, which both calculates height and removes the `hide` class, making `this.countryList` visible. To be more efficient, when an element is first painted on the screen, it is automatically painted in the correct position. So, rather than paint the element in one position, and then immediately reposition it, the browser is smart enough to know to do it all at once.

In the browser's eyes, instead of seeing:
```js
this.countryList.removeClass("hide"); // make visible

var that = this... // more code

this.countryList.css("top", cssTop); // position menu
```
it sees:
```js
this.countryList.removeClass("hide").css("top", cssTop); // position menu
```
Because `this.countryList` is a child of `this.dropdown`, and `this.dropdown` is absolutely positioned and initially contains only a hidden div, its visibility is entirely determined by `this.countryList`. So, even though `this.dropdown` is already appended to the page at line 5, it doesn't affect the layout at all, so lines 28-31 also happen at the same time. Essentially, the browser behaves as if the code is:
```js
//pseudo code
this.dropdown.css({"top": inputTop + extraTop, "left": pos.left}).find(this.countryList).removeClass("hide").css("top", cssTop); // position menu
```

caseyjhol:
Now using the `dropup` class instead of calculating cssTop for placing the menu above the input, as applying a class is a bit faster, and cleaner. I called the class `dropup` to keep it consistent with Bootstrap, but I can call it something different.

jackocnr:
Love the dropup `bottom: 100%` trick!

Thanks for all your hard work on this.

jackocnr:
Released v6.3.0 with this in it.

jackocnr:
So I did what you suggested and turned on "show paint rectangles" and I could see that Chrome was not rendering the dropdown below the input before it positioned it above. But I have a followup question: is this just clever optimisation by Chrome, or do all browsers do this? What about IE8? Thanks for your help.

caseyjhol:
As far as I know, it shouldn't be an issue. I also tested the paint rectangles in Firefox and didn't see anything. Even if it did place it on the page initially and then apply the class, the variables in between removing the `hide` class and applying the `dropup` class only take between 0 - 0.9 ms to calculate. It worked fine for me in IE8 in testing as well.

PR #259, opened 2015-07-28
Add option to prepend a custom message to the input's placeholder

efigarolam:
This PR adds a configuration option to customize the input's placeholder. For instance: "e.g.", this is illustrated on the following image: 

screen shot 2015-07-28 at 11 34 00 am

This can be done with the following code: 

````js
$('.js-phone').intlTelInput({
  utilsScript: '/js/vendor/libphonenumber-utils.js',
  prePlaceholderMessage: 'e.g. '
});
````


jackocnr:
Thanks for the idea - I will track it's popularity. But for now, I'm not going to add in extra options / LOC until I'm sure it would be useful for lots of people. For now, you can use the workaround I describe here (3 LOC): https://github.com/Bluefieldscom/intl-tel-input/issues/229#issuecomment-122603121

jackocnr:
FYI you can now use the `customPlaceholder` option to achieve this, as of v6.0.7.

efigarolam:
Nice, great job! it will be really helpful :smile:

Issue #260, opened 2015-07-31
How to update carrier metadata in utils.js from last libphonenumber?


xak2000:
I need last fresh metadata for carriers from last version of `googlei18n/libphonenumber`. It includes new carrier codes. How can I rebuild utils.js with this last metadata?

Specifically I need metadata from version 7.0.8.

xak2000:
I found an answer in https://github.com/Bluefieldscom/intl-tel-input/blob/master/lib/libphonenumber/src/utils.js 

Thanks for good documentation!

jackocnr:
Yes just bare in mind that libphonenumber (and a lot of its dependencies)
have moved to github, and last time I checked the docs hadn't been updated.

jackocnr:
Ok I updated the comments here: https://github.com/Bluefieldscom/intl-tel-input/blob/master/lib/libphonenumber/src/utils.js

And finally I see the libphonenumber docs have been updated, so you should be good to go! Let me know if you have any problems at all.

jackocnr:
(I would really appreciate it if you could do a pull request when you're done!)

xak2000:
I haven't any problems. libphonenumber's docs and docs from your utils.js is enought to do all I need in 5 minutes.

How can I do a pull request? Just fork your repo, change `lib/libphonenumber/build/utils.js` to new compiled utils.js and do pull request?

xak2000:
Just note: It will be good to have docs from utils.js in main readme.md file. I checked it and not found any documentation about recompilation utils.js. After that I opened this issue and only after that I guess to check utils.js code to find any js-docs.

jackocnr:
Yes you need to fork, update utils.js and do a pull request, thank you!

And that's a good idea re: the readme. Could you do that as part of the pull request? Maybe just add a sentence at the end of the Utilities Script section, saying where you can find the instructions.

jackocnr:
Hi @xak2000 - I have a big favour to ask - I've just noticed that libphonenumber have just published v7.0.9. Would you mind doing another build?

xak2000:
I recompiled utils.js with v7.0.9 and made a pull request.

jackocnr:
Thank you so much.

PR #263, opened 2015-08-03
Recompile utils.js from libphonenumber v7.0.8

xak2000:
See #260.


jackocnr:
Thanks a lot for doing this.

PR #281, opened 2015-08-28
Added EH,GG,YT,IM,JE countries dial code,  add two options disablePicker and disableUpdateFlagFromNumber and a public method isValidNumberForRegion

fishman1986:
Hi there,

First, thanks for the great job!
I met use this plugin in my project, and then I found some issues need to change the code. And I think this might be help.

* I noticed you remove some countries
>6f10b35 Jack O'Connor on 14-7-13 at 上午10:04 [intl-tel-input]
>removed lots of extra shit in data.js that wasn't included in google's tel input
But I find google support these for now.

* I met a requirement, it is a little like the example http://jackocnr.com/lib/intl-tel-input/examples/gen/country-sync.html . And I want to disable the country picker since there is already a select, and then, need to disable updateFlagFromNumber.

* I need to validate the phone number for region strictly, so I add a new public method isValidForRegion.

I don't know how to separate these commits into different pullrequest, but I think the first one not bad.


jackocnr:
Hi there, thank you for your suggestions!

Re: the extra countries, you say "But I find google support these for now" - but I just checked their telephone input on their gmail signup page and I don't see them.

Re: `disablePicker`,  I like this idea - it's a step towards solving issue https://github.com/Bluefieldscom/intl-tel-input/issues/248 - can we rename it `countryMode` and handle it like I describe in that issue? Also in this case we should not even generate the dropdown markup (you can comment it out in `_generateMarkup`). Also we will need to update the CSS because otherwise you will get hover effects that imply you can do something when you can't.

Re: `disableUpdateFlagFromNumber ` - I don't understand why we need this?

Then you mentioned adding `isValidForRegion` which I can't find anywhere in the code?

Finally: I'm going to close this issue because all of these should really be separate pull requests - one per feature. Then we can discuss them individually as needed.

Thanks again.

fishman1986:
Hi,

In this file, you can find these country code, support since Dec, 2014. And I'm sure it works fine.
https://github.com/googlei18n/libphonenumber/blob/6164bf22b822c8404b0741ac45cd7197d5bbb420/javascript/i18n/phonenumbers/metadata.js

*disableUpdateFlagFromNumber*, it is happened when I disable the country picker, for example, when user input '+86', the plugin will automatically update the country to China, which I don't want it happen.

About *isValidForRegion*, sorry, it is *isValidNumberForRegion*.

Anyway, I will find a way to create separate pull requests.

PR #283, opened 2015-08-31
Added EH,GG,YT,IM,JE countries dial code

fishman1986:
Hi there,

I fork this repository again, and create new pull request.

### DIFF
--- a/src/js/data.js
+++ b/src/js/data.js
@@ -482,6 +482,12 @@ var allCountries = [
     "502"
   ],
   [
+    "Guernsey",
+    "gg",
+    "44",
+    1
+  ],
+  [
     "Guinea (Guinée)",
     "gn",
     "224"
@@ -547,6 +553,12 @@ var allCountries = [
     "353"
   ],
   [
+    "Isle of Man",
+    "im",
+    "44",
+    2
+  ],
+  [
     "Israel (‫ישראל‬‎)",
     "il",
     "972"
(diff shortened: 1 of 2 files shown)


jackocnr:
Can you tell me why you want to include these areas? Do we even have flags for them?

Apparently Western Sahara is a disputed territory, so I don't want to include that. Guernsey, Isle of Man and Jersey are all subsideries of England, and have the same intl dial code, so I don't see the point in adding them. Remember this is all extra weight in the plugin and it's really important to keep this as small as possible.

fishman1986:
Can you tell me why you want to include these areas? Do we even have flags for them?
> Yes, we have.
According to the [ISO 3166-2|https://en.wikipedia.org/wiki/ISO_3166-2], they are officially assigned their own country codes.

jackocnr:
Ok I have added a section to [data.js](https://github.com/Bluefieldscom/intl-tel-input/blob/master/src/js/data.js) detailing the list of regions that I have omitted and a reason why.

If you would like to challenge any of these reasons, please state clearly why you think it is worth adding to the plugin, and give links to examples of other country lists which also include that region.

jackocnr:
You know what, I've changed my mind. If people from these regions identify that way, then we should support that.

jackocnr:
Oh it seems that you have deleted your repository now, so I can't just re-open this pull request - I'm sorry about that. Is it easy to get it back?

fishman1986:
New PR coming, https://github.com/Bluefieldscom/intl-tel-input/pull/295

PR #285, opened 2015-09-02
implementing custom placeholder function

ruhley:


jackocnr:
Hi there, thanks for this contribution. Can you please talk a bit about why you think this would be useful?

ruhley:
The placeholer currently has something like ```0412 345 678```. However some of our users were getting confused thinking that's what was already entered. With this I can make it say ```e.g. 0412 345 678``` which reduces the confusion.

jackocnr:
Thanks for this. Solves https://github.com/Bluefieldscom/intl-tel-input/pull/259 and could also be used to solve https://github.com/Bluefieldscom/intl-tel-input/issues/229.

rulzart:
Hey jackocnr

I have a question/problem. It's not directly related to this commit, but was the only way I could contact you (no PM system on Github anymore).

My issue is with jQuery UI Dialogs (modal) and the intl-tel input. When I have nationalmode set to false,  I need to click in the input field twice for it to get focus. With default settings (nationalmode: true), it works fine (1 click focus).

Sorry for writting in here. I really like your plugin - it is great!

/rulzart

jackocnr:
@rulzart please open a separate issue for this. Also I will need a codepen (or similar) showing the problem in order to help.

Issue #275, opened 2015-08-24
Support for Antarctica?


mpmurph:
Hi! I am trying to integrate an address form with collecting a user's phone number using intl-tel-input. 

I was randomly testing addresses and ran one for Antarctica which in turn populated intl-tel-input's defaultCountry with continent code "aq" and, when I got an error back, realized that intl-tel-input doesn't include Antarctica's dialing code of +672. 

I know this represents a tiny portion of users but was wondering if there are plans to include it? If the issue is establishing what to use for a flag, perhaps the emblem of the Antarctic Treaty (as officially resolved in 2002: http://www.ats.aq/devAS/info_measures_listitem.aspx?lang=e&id=289) could be used in lieu of a flag?

A minor point in an otherwise excellent package! Thank you!

mpmurph:
Just realized that +672 is also the dialing code for Norfolk Island and when I enter "nf" in lieu of "aq," the national number format for Norfolk Island seems to correspond with Antarctica's too (although I can't find any definite Antarctic numbers to be absolutely sure). In light of that, I will do a little behind the scenes country code switcheroo if a user enters an address in the Antarctic.

jackocnr:
According to Wikipedia's [List of country calling codes](https://en.wikipedia.org/wiki/List_of_country_calling_codes), +672 is the dial code of Australian External Territories, which include Norfolk Island (+672 3) and Australian Antarctic Territory (+672 1). I think we should update the metadata to reflect that, seeing as Australian Antarctic Territory just uses Australia's normal flag, so that's easy.

Unfortunately I don't have time to implement this myself right now, but I would be happy to accept a pull request.

jackocnr:
See comment on pull request.

PR #294, opened 2015-09-13
Added "aq" as Australian Antarctic Territory and updated Norfolk Island ("nf").

mpmurph:
Hi Jack - 

Apologies for the delayed follow-up on issue (#275). I have added "aq" as a code to the data.js file (and updated Norfolk Island - "nf" too). I saw your added note though at the top (about not including it) so am totally fine if you decide not to accept this pull request. I did not change the sprite.scss file as it looks like "aq" already exists there.

Side note: this is the first time I am doing a pull request so apologies if I am not doing it correctly! 

Rgds,
Mary

### DIFF
--- a/src/js/data.js
+++ b/src/js/data.js
@@ -15,7 +15,6 @@
 // UPDATE Sept 12th 2015
 // List of regions that have iso2 country codes, which I have chosen to omit:
 // (based on this information: https://en.wikipedia.org/wiki/List_of_country_calling_codes)
-// AQ - Antarctica - all different country codes depending on which "base"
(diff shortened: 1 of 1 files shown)


jackocnr:
Thank you for taking the time to do this, but I have realised that it doesn't make sense to add any single country calling code for antarctica because according to this [List of country calling codes](https://en.wikipedia.org/wiki/List_of_country_calling_codes#Locations_with_no_country_code) - if you search the page for "antarctica" you will see that there are lots of different calling codes depending on which "base" you chose in the area.

mpmurph:
No worries - thanks for responding!

PR #301, opened 2015-09-23
Remove jQuery dependency

Ahimta:
The library now works without jQuery using native Javascript.
jQuery is now only a development dependency for tests using
jasmine-jquery matchers.

However, the jQuery plugin is still added if jQuery is defined.

Some caveats:
1- I am not very confident of src/js/wrapper.js.ejs correctness.
2- The minimized version produces errors when jQuery is not defined
  while the non-minfied version works correctly. This is probably
  because of some hacky behavior in the tooling. Most likely in
  grunt/replace.js. In addition, appearantly, replace tasks aren't
  applied to all files (e.g: some files spec/). Anyways, I personally
  think that grunt/replace.js should be replaced by a more
  conventional solution; if library size is so important then all
  scripts that are always used should be concatenated, uglified and
  gzipped.

Please, review and let me know what I should do to make it easy to merge the PR.

This PR is only one commit squashed from [here](https://github.com/Ahimta/intl-tel-input/tree/abandon-jquery)
Almost, every single commit in the original branch is functional, however, the commit messages aren't all that helpfull.

You can find the demo here: http://ahimta.github.io/intl-tel-input/


jackocnr:
Wow this is super cool, thank you! I had wanted to look into this anyway, but not had the time.

My only concern is filesize, and it looks like there's a lot of green additions in this pull request! Do you have an idea of filesize of the built minified file before and after this change?

Ahimta:
Before: 23 KB
After: 30 KB

Unfortunately, for now the minified version doesn't work without. Most likely grunt/replace.js has something to do with it.

I didn't have filesize in mind when I was working on this. If you really care about filesize that much, I can make it a little bit smaller.

However, I think the most important part is to concatinate all files that are always used, uglify them and then gzip them. In general trying to make it smaller, more using tools, than manual effort.

jackocnr:
Hmm have you read this doc? https://twitter.com/paul_irish/status/431584056883429376

jackocnr:
Seems like a complete nightmare, keeping up with all the different bugs for the different browsers, and updating all of the extra checks/hacks as they are fixed, or as we drop support for old browsers etc...

Ahimta:
I would almost never use the native Javascript API for web apps. I would use (e.g: Angular, React).
Even writing a web app using jQuery is time consuming.

However, for Javascript libraries it's best to not use jQuery (and many Javascript libs don't use jQuery).
In reality, not everyone is using jQuery and if they do, they are using different versions that could produce different behavior and break the library code.

I read the [doc](https://docs.google.com/document/d/1LPaPA30bLUB_publLIMF0RlhdnPx_ePXm7oW02iiT6o).
jQuery has many many features and intl-tel-input is using only a small subset of them, that why we can without much effort make it work without jQuery.

Additionally, [jQuery browser support](http://jquery.com/browser-support/) isn't as impressive as most people would think.
And I am happy that it isn't. Otherwise the development would be crippled and we would end up with a worse product.

In my opinion, browser support is a tradeoff. If you want to support older browsers (e.g: IE7, IE8) there's a big price (even jQuery2 doesn't support them). It's okay to do it as long as you understand the tradeoff.

jackocnr:
> In reality, not everyone is using jQuery and if they do, they are using different versions that could produce different behavior and break the library code.

We work with >=1.7 which was released in 2011, so unlikely to be a problem.

> jQuery has many many features and intl-tel-input is using only a small subset of them

I disagree. We're using:
- selectors
- element creation
- DOM manipulation
- event listening
- $.extend
- $.Deferred
- $.inArray
- $.isNumeric
- $.trim
- $.ajax
- $.data
- $.when
- $.each

All of these contain their own fixes for various browsers/versions (even modern browsers), which we would now have to manually keep up-to-date. We also lose out on the millions of people using jQuery providing bug reports, and the fixes that get written in response.

> Additionally, jQuery browser support isn't as impressive as most people would think.

I'm not sure what your point is - if you look at that doc it is clear they have fixes for hundreds of browser bugs. If we don't use jQuery we have to do all of that ourselves.

> If you want to support older browsers (e.g: IE7, IE8) there's a big price

I hate supporting these old browsers, but it's fairly easy to do in this case - just use jQuery v1.

I'm sorry but I don't think we should drop jQuery for two reasons: stability (in terms of browser support), and filesize.

Ahimta:
Thanks for your honesty and fast reply, seriously!

I think it's worth it for Javascript libs to work without jQuery.
So I will work on my fork for the long term; that's the beauty of open source.

What I plan to do (mostly):
1- Maintain my own fork.
2- Incorporate all new intl-tel-input features into my fork.
3- Make sure it has a good browser support, probably using [https://saucelabs.com/](Sauce Labs).
4- Move jQuery plugin to its own repo (using my fork through package mangers: npm, bower, component).
5- Write different wrappers, each in its own repo for Angular, React, Ember (eventually).

jackocnr:
Sounds good. Thanks for your interest in the plugin!

se-panfilov:
Thanks, for work, it's perfect widget.

But, you know, jQuery really make sense for project based on other framework (Angularjs, for me as example)

If I want use your perfect lib, i should make 3 steps:
1. Include angular wrapper (like https://github.com/mareczek/international-phone-number);
2. Include your lib;
3. Include jQuery;

It's a bit overkill for me, cause angularjs already has jqlite inside and no other libs required for jquery.

I want to ask - is there any known forks without jQuery dependency?

@Ahimta How works going on? Is it your repo stable?

Ahimta:
@se-panfilov you can use https://github.com/Ahimta/bc-phone-number.

It depends only on Angular, Angular UI, Bootstrap CSS files (not js files) and some other libs I made
specifically for bc-phone-number.

jackocnr:
Update: I'm now looking to try to move away from jQuery incrementally - help appreciated! https://github.com/jackocnr/intl-tel-input/issues/405

Ahimta:
You can start by refactoring. For example, factor out the logic that doesn't depend on the DOM into its own functions, and have the parts that deal with the DOM, events, etc... call these pure functions.

PR #302, opened 2015-09-24
Added the last countries

fishman1986:
Added some can be supported countries, update the comment.
https://github.com/Bluefieldscom/intl-tel-input/issues/297


jackocnr:
Thanks for this - I have manually merged it in.

Issue #300, opened 2015-09-23
RTL Support


Francisc:
Hello,

I needed to add RTL support for this plugin in a project where I use it.
I would like to send a PR that enables RTL support for it.
This works like this:
You either set a new config option `rtl: true` or set a class on any ancestor. The option does that as well, sets a class.
The result is the flag is displayed on the right and phone number is right-aligned.

Let me know.

jackocnr:
Hi there, yes that sounds great thanks - please send a pull request so I can have a look. I think we should just go for a class for now, as I'm guessing this only effects the CSS, not the JS? Shall we call it `.iti-rtl`?

iamstarkov:
made few comments in pull-request

jackocnr:
Closing due to inactivity.

DannyBen:
Sorry to resurrect this, does it mean that there is no RTL support yet?

jackocnr:
@DannyBen that is correct. If you read the PR comments, you will see that the implementation was never finished properly. I would be open to a fresh PR tho!

DannyBen:
I actually found a simpler, and arguably better solution.

Since a) the dropdown list is always in English (+ localized name), and b) phone numbers should be written Left to Right in all languages, then it actually makes sense to simply wrap the input control with `<div dir='ltr'>`. When this is done, all is good with the world again.

I think if you just mention this in the README somewhere, most users who require RTL will be satisfied.

My slim code (on Rails), for reference:

```slim
.field.col.span-6
  = f.label :phone
  div dir='ltr'
    = f.telephone_field :phone, dir: 'ltr', class: 'align-left'
```

And on that note, I want to say this plugin works flawlessly. I was worried it will mess up my CSS style, ot look weird in most situations, but it looks very naturally. Well done.

yehudahkay:
Thanks! Theoretically it would be good if the div that is created to wrap the input would have dir='ltr'

jackocnr:
There is now a RTL demo page here for future reference: https://intl-tel-input.com/examples/right-to-left.html

PR #308, opened 2015-10-07
RTL support.

Francisc:
A note about `build/css/intlTelInput.css` although originally it was on a single line, the Grunt style config of `compressed` outputs it as can be seen in the PR.

Issue #300.


src/js/intlTelInput.js.ejs
```diff
@@ -246,6 +250,11 @@ Plugin.prototype = {
       // this is useful in lots of places
       this.countryListItems = this.countryList.children(".country");
     }
+    
+    // add the RTL class if requested
+    if(this.options.rtl === true) {
+    	this.telInputRootElement.addClass("iti-rtl");
```
jackocnr: no need to have this cached if we're only going to use it once - just get the parent here instead

src/css/intlTelInput.scss
```diff
@@ -229,6 +230,41 @@ $flagsImageExtension: "png" !default;
...
+      
+      .iti-flag {
+        left: $flagPadding;
+      }
+    }
+    
+    .country-list {
+      .flag, .country-name {
+        // cancel right margin and set it as left margin
+        margin-left: 6px;
```
jackocnr: if we're going to use this number twice, let's use a variable

jackocnr:
Don't worry about the CSS build being weird - I can fix that. You will need to fix the conflicts though (since I merged in another pull request).

src/css/intlTelInput.scss
Francisc: Done.

src/js/intlTelInput.js.ejs
Francisc: It's only used once and is the direct parent now.
Later that may change and it could potentially break the RTL styling by placing the class on the wrong element.
Also, it's a good idea to keep reference to the top-most element of the component and introduces no performance issue.
I'd keep this as is.

jackocnr: Ok, we also use it in the `destroy` method. If we're going to do this, we should do it properly. It should be set in `generateMarkup` when the root element is created, and let's call it `this.rootElement`. Please also update the `destroy` method to use it too. Thanks.

build/css/intlTelInput.css
```diff
@@ -1 +1,933 @@
...}.iti-flag.vi{height:14px;background-position:-5390px 0px}.iti-flag.vn{height:14px;background-position:-5412px 0px}.iti-flag.vu{height:12px;background-position:-5434px 0px}.iti-flag.wf{height:14px;background-position:-5456px 0px}.iti-flag.ws{height:10px;background-position:-5478px 0px}.iti-flag.xk{height:15px;background-position:-5500px 0px}.iti-flag.ye{height:14px;background-position:-5522px 0px}.iti-flag.yt{height:14px;background-position:-5544px 0px}.iti-flag.za{height:14px;background-position:-5566px 0px}.iti-flag.zm{height:14px;background-position:-5588px 0px}.iti-flag.zw{height:10px;background-position:-5610px 0px}.iti-flag{width:20px;height:15px;box-shadow:0px 0px 1px 0px #888;background-image:url("../img/flags.png");background-repeat:no-repeat;background-color:#DBDBDB;background-position:20px 0}@media only screen and (-webkit-min-device-pixel-ratio: 2), only screen and (min--moz-device-pixel-ratio: 2), only screen and (-o-min-device-pixel-ratio: 2 / 1), only screen and (min-device-pixel-ratio: 2), only screen and (min-resolution: 192dpi), only screen and (min-resolution: 2dppx){.iti-flag{background-image:url("../img/flags@2x.png")}}.iti-flag.np{background-color:transparent}
```
iamstarkov: is it possible not to include auto-generated files in pull-request, it makes it closely to impossible for reviewing

src/css/intlTelInput.scss
```diff
@@ -229,6 +234,41 @@ $flagsImageExtension: "png" !default;
...
+        margin-left: $countryListSpacing;
+        margin-right: 0;
+      }
+      .dial-code {
+        // dial codes should be LTR
+        direction: ltr;
+      }
+    }
+  }
 }
```
iamstarkov: manual process is really error prone, what do you think about using rtlcss here?
https://github.com/MohammadYounes/rtlcss

build/css/intlTelInput.css
Francisc: The minified file was already there, so not updating it seemed wrong.

src/css/intlTelInput.scss
Francisc: Not sure how this could be automated to be honest.
It's very custom, so to speak.
If you believe the project you linked could do this, then great, close the PR unmerged and do it.

build/css/intlTelInput.css
iamstarkov: from my perspective it should be done by maintainers, once they will make new release

jackocnr:
@Francisc I just created [a codepen](http://codepen.io/jackocnr/pen/YwqabW) to test out your CSS and I found a few issues/questions:

1. I see a couple of `direction:ltr` statements in there - can you explain these? I notice that when you type the phone number, you're still typing LTR - does this make sense?
2. Some country names containing parenthesis get all messed up - see Afghanistan. I think it might be caused by languages that are normally RTL.
3. In the dropdown there's no spacing between the dial code and the country name. When you cancel the margin-right on the flag and the country name, and you add the margin-left instead, you need to add it to the country name and the dial code instead, due to the ordering.

Also, see my comment about `this.rootElement`.

jackocnr:
Closing due to inactivity. If anyone wants to pick this up again, please let me know!

Francisc:
1. Phone number and numbers in general are supposed to be displayed LTR.
2. We need to add a control character or try CSS `unicode-bidi:embed;`. http://www.microsoft.com/middleeast/msdn/control.aspx
3. That's odd, I thought this was fixed before submitting the PR.

jackocnr:
Hi @Francisc - it's good to hear from you again. Shall I re-open the pull request? Are you going to be working on this again?

Francisc:
Hi, time's a problem for me.
I'll try to address points 2 and 3 from the above list.

PR #310, opened 2015-10-12
add excludeCountries option (close #200)

### LINKED ISSUE #200: Feature request: exclusion list
Hello, this is an amazing plugin, and I saw you already have `onlyCountries` option. It would be also nice to have something like a `excludeCountries` option to do the opposite, setting which countries we don't want to show.

Thanks.

caseyjhol:
If onlyCountries and excludeCountries are both set, excludeCountries is ignored.

See example here: http://plnkr.co/edit/aSqUSWRenMmhE6KcrlR8?p=preview


jackocnr:
Haha you don't stop! Thanks for putting this together. I will review properly ASAP. I apologise that I've been very busy with work recently (and continue to be), but at a minimum I try to give this plugin some time on the weekends.

jackocnr:
This looks great - thanks again!

PR #321, opened 2015-11-10
Switching to js-cookie (since jQuery-cookie is now defunct).

### LINKED ISSUE #320: Switch to JS Cookie
Now that jquery cookie is defunct, any plans to switch to JS Cookie? If you're OK with the idea, I can help send a PR for the change.

mrchief:
Fixes #320.


src/js/intlTelInput.js.ejs
```diff
@@ -429,7 +429,7 @@ Plugin.prototype = {
     var that = this;
 
     // check for cookie
-    var cookieAutoCountry = ($.cookie) ? $.cookie("itiAutoCountry") : "";
+    var cookieAutoCountry = (!!Cookies.get) ? Cookies.get("itiAutoCountry") : "";
```
jackocnr: If Cookies is not defined this will error. Instead you should check if (window.Cookies)

jackocnr:
Thanks for this. I added a comment in the code. Also could you please update the reference in the readme?

src/js/intlTelInput.js.ejs
mrchief: Yeah, realized after I committed. Let  me make the changes (readme as well).

src/js/intlTelInput.js.ejs
```diff
@@ -429,7 +429,7 @@ Plugin.prototype = {
     var that = this;
 
     // check for cookie
-    var cookieAutoCountry = ($.cookie) ? $.cookie("itiAutoCountry") : "";
+    var cookieAutoCountry = (!!window.Cookies) ? Cookies.get("itiAutoCountry") : "";
```
jackocnr: Can you tell me the reason for the double bang? I know it casts to Boolean, but that happens automatically in conditionals like this. Is it somehow safer? Or do you think it is more readable? Because I think it's less readable. (it's also inconsistent with the rest of the codebase).

mrchief: Its the same difference as in `==` vs `===` (or `!=` vs `!==`).

Without these, what you're comparing is truthy/falsey which works fine for cases like this, but readability-wise, I think this conveys your intent better.

Again, JS is a highly opinionated language and you're the author and this isn't a battle worth fighting; so if you want to change it, please go ahead! :)

jackocnr: Ok thanks for the clarification. I actually disagree that this is more readable - I think it's obfuscated and confusing. Perhaps doing an explicit cast to boolean would be more readable. Either way, I think we should go for consistency for now. Also please can you run `grunt js` to build the JS files?

jackocnr:
Thank you very much!

PR #325, opened 2015-11-19
Added setter for countries data

matteoantoci:
I added the public setter for countries data


jackocnr:
We actually had this before, but I removed it because you can do everything with the getter. If you first use the getter, and then modify that object, then those changes are reflected in the plugin.

jackocnr:
As per my last comment: unless you have a real use case that justifies this extra weight, then I don't see a reason to add it.

matteoantoci:
Sorry, for the late reply. No problem, thanks!

PR #334, opened 2015-12-12
Don't assume remainder is an extension if the result is still too short.  Fixes #329.

### LINKED ISSUE #329: Extn parsing issue
I have an otherwise valid number that is failing to format/validate when I enable extension support.  It looks like the as-you-type output is being misinterpreted and an extension is being assumed prematurely.

Using the [libphonenumber demo](https://rawgit.com/googlei18n/libphonenumber/master/javascript/i18n/phonenumbers/demo-compiled.html) with 55226... as the national format number and Chile (CL) as the country...
> ****AsYouTypeFormatter Results****
> Char entered: 5 Output: 5
> Char entered: 5 Output: 55
> Char entered: 2 Output: 55 2
> Char entered: 2 Output: 5522
> Char entered: 6 Output: 55226
> Char entered: 6 Output: 55 226 6

Notice how the space goes away for a couple of characters and then comes back.  In the `formatNumber` function in utils.js, there's a loop that sets the extension flag.  I've stepped through this code (which is tricky since there isn't an unminified version of utils.js included), and it's evident that the length change when the space goes away is what trips the [extension condition](https://github.com/jackocnr/intl-tel-input/blob/master/lib/libphonenumber/src/utils.js#L103).  If I type `5522`, it ends up being formatted as `55 2 ext. 2`.  ([demo](http://codepen.io/anon/pen/zrGBqY))

The sudden shrinkage when the spaces the formatter adds disappear is a hint that it's time to switch over to the extension, but it looks like the logic should keep going to see if the formatter starts adding spaces again later on.  But if the user is typing the number, then you aren't going to have more digits to process, and the user is going to see the format switch to the extension mode and then switch back once enough digits are typed that the format is recognized as valid.

In the libphonenumber demo, it reports `Result from isPossibleNumberWithReason(): TOO_SHORT`.  I think that may be the only reliable way to detect when it's time to switch over to the extension.  The length/space based check isn't universal enough.

nlwillia:
Oops...PR'd to my fork instead of your base.  Trying again.


jackocnr:
Love it, thank you.

PR #340, opened 2015-12-28
Fixed javascript bug on the is-valid-number example

wcyn:
Fixed the below issue -  the validation information is appended when another country is selected from the dropdown. 
screenshot from 2015-12-28 16 51 30

### DIFF
--- a/examples/js/isValidNumber.js
+++ b/examples/js/isValidNumber.js
@@ -2,8 +2,18 @@ var telInput = $("#phone"),
   errorMsg = $("#error-msg"),
   validMsg = $("#valid-msg");
 
-function validate(){
-  //display validation 
+// initialise plugin
+telInput.intlTelInput({
+  utilsScript: "../../lib/libphonenumber/build/utils.js"
+});
+
+
+telInput.on("change", function() {
+  // reset validation
+  telInput.removeClass("error");
+  errorMsg.addClass("hide");
+  validMsg.addClass("hide");
+
   if ($.trim(telInput.val())) {
     if (telInput.intlTelInput("isValidNumber")) {
       telInput.removeClass("error");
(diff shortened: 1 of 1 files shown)


jackocnr:
Hi Cynthia,

Thank you for reporting the issue. I would actually do the fix slightly differently, because the original intention of the code was to A) only validate on blur (so it's not constantly saying "invalid" while you're still trying to type your number), and B) try to clear any existing validation messages when the user modifies the number.

So the way I would fix it would be to leave the validation code in the blur function, and instead extract the reset code into it's own function, and call that at the start of the blur function. Also the plugin triggers a "change" event when the user selects a different country, so if you change the keydown listener to the following: `telInput.on("keyup change", function() {` then that will catch both events (so it will then reset any validation messages both when they change the number and when they change the country).

Does that make sense?

wcyn:
I see what you mean.
I've made some changes to the example file. I've found that the change method is enough to handle the blur events. The form is validated only when the input loses focus, and when a different country is selected. Adding the keyup event defies A, so I've left that out.

jackocnr:
Thanks again for your input, but unfortunately there are still some issues. You say "the change method is enough to handle the blur event", but actually we want to handle the two situations differently: when you change the flag we just want to clear any validation messages, but when you blur the input, then we want to display the validation message. Also you say "the keyup defies A", but I disagree - in the keyup handler we don't do any validation, we just remove any existing messages. If we omit the keyup handler, it will display the old validation message while the user types a new number, which is confusing.

For these reasons, I have just implemented my own solution, but thanks again for raising the issue.

wcyn:
Ah, yes. I missed that bit. Now I see exactly what the code was meant to do. Hope to be more helpful next time

PR #347, opened 2016-01-19
Ignore all inside lib folder except libphonenumber

oleksiireshetov:
After running of the ```grunt bower``` command there is the *region-flags* folder inside the *lib* folder that is not ignored. The *.gitignore* file was updated to ignore all files and folders inside the *lib* folder except  the *libphonenumber* folder.


jackocnr:
Thanks!

Issue #317, opened 2015-11-04
Show Country dial code instead of flag


rockom:
Hey
Would be great to have the option of to show in the dropdown the country dial code instead of the flag.

Thank you,
Roy

jackocnr:
Hi Roy, can you tell me why you'd like to do this?

nickovruchsky:
+1
This is really useful feature. For now my client want no flags but only code. So I should implement this by myself I think.

rockom:
Hi,

I have the same issue.
I have clients who prefer flags and clients who prefer dial code only.

Thank you,
Roy

jackocnr:
Why don't you just hide the flags using CSS? The dropdown already shows dial codes.

rockom:
I am not sure it looks good but I will try to do so.

nickovruchsky:
The main idea is to show the codes in input while it unselected.

rockom:
Hiding the flags using CSS provide bad user experience.
It is much better to show the dial code or flags based on the options.

caseyjhol:
@rockom Not quite sure what you're referring to. Can you make a mockup in Photoshop of what you want the menu to look like?

```css
.intl-tel-input .country-list .iti-flag {
    display: none;
}
```

Seems to accomplish what you're asking. Then, set `nationalMode: false`.

nickovruchsky:
For me it's look like this - 
![tel](https://cloud.githubusercontent.com/assets/7410355/10981022/be3db878-840d-11e5-9558-a6e4a7aed321.jpg)

rockom:
@nickovruchsky , This is exactly what i am looking for.
@caseyjhol , When applying your css i get:
![number-dropdown](https://cloud.githubusercontent.com/assets/3122712/11019148/859746aa-85f3-11e5-92ab-15fcc239617c.png)

nickovruchsky:
@rockom try this 
```scss
      .intl-tel-input { 
        .form-control {
          padding-left: 25px;
        }
      .flag-dropdown {
        .selected-flag {
          width: 20px;
        }
        .iti-flag {
          background: none;
          .arrow {
            left: 0; 
          }
        }
      }
    }

jackocnr:
[...]

PR #349, opened 2016-01-20
i18n support & hide/show flags

john-e:
I have added locale customization to the plugin, as i needed in one of the projects. And also an option to show or hide flag in list/next to input box


jackocnr:
Hi @john-e, thanks for all this work! So there are lots of new features in here to discuss.

Firstly, the localization - I see you have added German, French and Spanish translations of all of the country names to the data.js file, and you've also added new options: `locale` to specify which language to use and a `translations` object to hold multiple translations. Now I'm not very experienced with localization, but it doesn't seem like the right approach to include a handful of languages in data.js by default, because that adds a lot of weight to the plugin and there are lots of devs who A) are not interested in localization yet, or B) who's users don't speak those specific languages, so it would be pointless for them to include them in the core plugin.

The same applies to the `translations` option - I don't think it makes sense to allow devs to pass in multiple languages at once. I think if you know the locale of a user, you should just pass in the translation for that one language. So I would propose we remove the translations from data.js, and we alter the `translations` option to just take one language. What do you think?

jackocnr:
Next, the `selectedShowFlag` option, which decides if you should show the selected flag or the dial code to the left of the input. Unfortunately there are issues with this with countries where national numbers contain extra digits to international numbers e.g. the UK. See my comment here: https://github.com/jackocnr/intl-tel-input/issues/317#issuecomment-156264036

This is a popular feature request though, so if you were willing to issues I mentioned, then I would welcome a separate pull request for that.

And finally, re: the `showFlags` option, I don't see the need for a new option when you can just do `.flag-box {display: none;}` in your CSS. I'm really committed to keeping this plugin as lean as possible.

john-e:
Yeah, it makes sense :+1: 

- `translation` can have one language set.

- For showing dialing code though, we can show them, as 0 in the UK, even in Indian number, it can be omitted as the number input has to merge with the dialing code, otherwise the plugin wont make any sense, 0 is given as the base country code (which country you're in), say if i dial 091234 56789 in India, and the number belongs to different state than caller, it will connect to it. But if i call the same number without 0, from different state, it wont connect. But if i add the +91 dialing code, like so +91 91234 56789
call it from anywhere in the world, it would connect.

inikoo:
About the translations how about passing the data by ajax from a external script

pellejacobs:
Hi @jackocnr, 
I was actually looking for a functionality to not load the countries, instead of just not displaying it. As for the moment, this plugin messes with our keywords: we are ranking for words such as 'islands' and 'guinea'. 
Cheers,
Pelle

jackocnr:
Hi @john-e,

Yes I understand that we must merge the dial code with the number the user types - that is my point - in any functions where we do something with the number (e.g. formatting) we must modify them to first add the dial code to the number, then do the formatting, then remove the dial code again before we insert the formatted number back into the input. Does that make sense?

Either way, I'm going to close this pull request as A) it needs some work anyway, and B) it contains multiple new features - in the future please open one pull request per feature. So one for the new `translation` option (no need for `locale` option if there's only one translation), and a separate one for the new `showSelectedDialCode` option, which should show the selected dial code next to the selected flag.

Thanks!

jackocnr:
@inikoo @pellejacobs I think the perfect solution to this translation issue is to remove the English country names from data.js and then have a separate language file which devs must include as well, but this seems like a lot of work, and while I would welcome a pull request that did that, I think that @john-e's solution is a great quick fix that will allow people to use this plugin in any language.

Issue #346, opened 2016-01-18
Disable and remove autoFormat feature


jackocnr:
I have been back and forth about this a lot - I spent a lot of time implementing the `autoFormat` feature and fixing bugs, and I think it's a super cool feature, BUT I think it's time to disable and remove it for the following reasons:

- First and foremost, we are hacking [libphonenumber](https://github.com/googlei18n/libphonenumber) to get this to work, which is dangerous. It was not made for this purpose, they do not support our use case, and they have [explicitly advised us against using it this way](https://github.com/jackocnr/intl-tel-input/issues/322#issuecomment-170599919).
- Last time I checked, `autoFormat` didn't work in Android Cordova apps, or in IE Mobile (see Browser Compatibility section of readme). Who knows what other browsers/versions/platforms make it not work properly.
- There is no way around this incredibly counter-intuitive UX issue: https://github.com/jackocnr/intl-tel-input/issues/322 where you cannot type the first character from the placeholder. This affects the USA and most likely lots of other countries too.
- Another minor issue: when we fix the above issue, it will lead to a new one: when a country placeholder starts with a formatting char (e.g. US starts with an open bracket), if you press that char to try and start the number, you will get a red flash as that is not an accepted key, but then it will still add the char because autoFormat does a lookahead and adds any upcoming formatting chars on each keypress, which leads to a confusing experience.
- Then there's all the issues we have already come across, and those we don't yet know about...
- Finally, restricting what the user can do in a regular `<input>` is considered bad practice by a lot of people, and as we have seen is very prone to bugs, and when there are bugs it can completely ruin the UX, or even render the whole input useless.

Notes
- We should still keep a `formatNumber` function available, which should be used with `setNumber` etc.
- This change would render the `allowExtensions` option useless, so we would remove that too.
- This change would also close issues: https://github.com/jackocnr/intl-tel-input/issues/190, https://github.com/jackocnr/intl-tel-input/issues/262 and https://github.com/jackocnr/intl-tel-input/issues/322
- Another perk is that this should significantly decrease the file size of the plugin.

I'm open to discussion, but at this point I'm pretty convinced, and would welcome a pull request that implemented this change.

nlwillia:
I should preface this by acknowledging that I am not a domain expert on international phone numbers.  I had not been exposed to E.164 before encountering this library.  So what I assume to be sensible about how to deal with input issues comes from an engineering perspective, feedback from my own users and what experience I have with numbers in my own nationality.

[...]

PR #364, opened 2016-02-12
isValidNumber now returns null value when utils script is not available

### LINKED ISSUE #362: isValidNumber assumes that number is invalid
Hello!

The `isValidNumber` API method uses `intlTelInputUtils.isValidNumber` function to actually validate the phone number. However, when `utils.js` is not loaded, it [assumes that number is incorrect](https://github.com/jackocnr/intl-tel-input/blob/master/src/js/intlTelInput.js.ejs#L1070).

I think that library shouldn't make such assumptions without any base. So in case when `utils.js` is not available it should fallback to some other internal validation method, e.g. simple RegExp or just return `null` instead of `false`.

I believe that `null` is more correct result when you can't came up with the actual result.

What do you think? Thank you.

slavafomin:
Fixes #362.


jackocnr:
Thanks!

slavafomin:
My pleasure. Will you be able to release this change in the new version soon?

jackocnr:
Actually v8.0.1. Done.

slavafomin:
Oh, I would really want to use `autoFormat` feature. This is probably the most important thing in the whole library for us. It works pretty great with international phone numbers in our locality. Are you planning to completely drop support for this feature?

I'm going to stick with the `7.x` branch for now then. Could you please merge this change into this branch as well?

jackocnr:
I feel your pain, but unfortunately I had to remove `autoFormat` because it is fundamentally broken for 49 countries including USA. See a full list here: https://github.com/jackocnr/intl-tel-input/issues/346#issuecomment-183771143. If you want your site to work for people from those countries, then you shouldn't use `autoFormat`.

If you still choose to, you can continue to use v7 of the plugin, but I will not be maintaining it.

PR #377, opened 2016-02-22
Make mobile popup position fixed

dtreiter:
Before, when the mobile popup was opened from the bottom of a long page it could appear too high up, preventing it from being visible.

Using a fixed position ensures the popup will always be visible in the same place relative to the users viewport, not just the top of the page.

Seems like a quick change, but are there any unexpected things to look out for with this?


jackocnr:
Makes sense, thank you!

PR #393, opened 2016-03-15
Disp;ay of Country prefix

kingsleyudenewu:
This event will display the country prefix into the input fields all
they have to do is to click on the country flag and the prefix will be
added automatically to the input fields.


jackocnr:
You can just disable `nationalMode` if you want this behaviour.

Issue #401, opened 2016-04-14
autoHideDialCode only works on blur, but should also work on form submit


ehmicky:
The option `autoHideDialCode` empties the input value on blur if it is invalid (e.g. if it is `+46`). However if the input is focused and the form is submitted without blurring (e.g. by hitting `Enter` key), the value is not emptied.

### Steps to reproduce
0.  Put the intl-tel-input widget inside a form
1.  Insert `+46125`
2.  Blur
2.  Edit to `+46`. Do not blur
3.  Submit form by hitting `Enter` key

### Expected behaviour
The input value should be emptied.

### Actual behaviour
The input value is not emptied.

### Initialisation options
Using default options.

### Possible fix
https://github.com/jackocnr/intl-tel-input/blob/master/src/js/intlTelInput.js.ejs#L552
The event handler here should be run not only on blur, but also `this.telInput.prop("form").on("submit")`, if form exists.

jackocnr:
Good idea - I would welcome a pull request that does this!

ehmicky:
Added a pull request at #404

jackocnr:
Released in v8.5.0 thank you!

PR #404, opened 2016-04-18
autoHideDialCode only works on blur, but should also work on form submit

ehmicky:
See #401


jackocnr:
Thank you very much for this! I didn't know about the form property on input elements - that's useful! And thanks for adding a test as well - great!

A couple of issues:
- we need to namespace the submit event, the same way we do with the blur event, to avoid clashes
- also please call `.off()` in the `destroy` method - again, the same as we do with blur.

Looking forward to merging it in :)

ehmicky:
Updated :)

PR #410, opened 2016-04-27
Fix typo in README

mpospelov:
Fix type and put the example with formatting argument.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,9 +234,11 @@ $("#phone").intlTelInput("setCountry", "gb");
 ```
 
 **setNumber**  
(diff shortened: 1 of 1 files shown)


jackocnr:
Thank you for this. I've actually just simplified the logic around setNumber, so now you actually don't have to pass a format enum - instead it automatically will format the number according to the nationalMode and separateDialCode options.

I'm interested in feedback tho - will this change work for your use case?

mpospelov:
I think no. I caught a bug in my application, I use angular directive https://github.com/mareczek/international-phone-number and for my case I have to store 2 values one in international format(ng-model) and one the original input value. The international format value can be changed from outside of directive and it cause the original input value to change the format. To prevent this issue before changing I'm checking the current format and use this `setNumber` method. I think for my special case it won't work so I have explicitly pass the format.

jackocnr:
I'm sorry I don't understand. Please can you explain in more detail, and with an example?

jackocnr:
Closing due to inactivity.

PR #412, opened 2016-05-04
Add option to allow placeholder overwrite.

jtwalters:
This PR adds an option to allow an existing placeholder attribute to be overwritten.


jackocnr:
Hi there, thanks for this. Can you please explain your use case? Why wouldn't you just not set a placeholder in the first place?

jtwalters:
Because:
- There's some delay in loading the utils script, etc., with the asynchronous nature of the way it's loaded, it results in the placeholder being blank for 1-2 seconds in my case. (poor experience)
- It seems like a valid option you should provide to folks, and doesn't require much additional code (+2 lines)

jtwalters:
Oh, I did see a reference: 

> If there is already a placeholder attribute set on the input then that will take precedence

Sorry. Missed that. But, my use case remains: I would like to avoid a blank placeholder while the utils script is loading.

jackocnr:
Ok fair enough, but I really don't want to add a third initialization option to this plugin just for the placeholder. Instead, let's adapt the existing `autoPlaceholder` option to now be a string, with 3 possible values: "off", "polite" (default) and "aggressive". Does that make sense?

jtwalters:
Sure, makes sense, I can rewrite it.

jackocnr:
I'm going to close this for now. Cheers.

PR #414, opened 2016-05-06
fitWidth feature added

vitaliy-patsay:
Created option to set countries list dropdown the same width  as initial
input width

### DIFF
--- a/src/css/intlTelInput.scss
+++ b/src/css/intlTelInput.scss
@@ -179,6 +179,7 @@ $mobilePopupMargin: 30px;
     }
 
     max-height: 200px;
+    overflow-x: hidden;
     overflow-y: scroll;
 
     // the divider below the preferred countries
--- a/src/js/intlTelInput.js.ejs
+++ b/src/js/intlTelInput.js.ejs
@@ -14,6 +14,8 @@ var pluginName = "intlTelInput",
     dropdownContainer: "",
     // don't display these countries
     excludeCountries: [],
+    // set countries list dropdown width the same as initial input width
+    fitWidth: false,
     // format the input value during initialisation
     formatOnInit: true,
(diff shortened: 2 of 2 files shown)


jackocnr:
Hi there, thanks for this. We actually used to have an option that did this called `responsiveDropdown`, but we dropped it when we figured out how to set the dropdown width dynamically based on the contents (so now it's only as wide as the longest country name is, if you get what I mean). I feel like this (width based on it's content) is a better approach. If we force it to be the same size as the input, then most likely you will have a lot of country names spilling onto 2 lines.

Can you please explain your use case?

vitaliy-patsay:
Hi,

I had exact use case for this option. It causes horizontal scroll on jQuery-UI dialog. Here is a [screenshot attached](http://content.screencast.com/users/scorpioghost/folders/Jing/media/c45f6017-fc00-47be-8113-6bfc9718fe38/2016-05-04_1343.png). And it's better to align dropdown with input for this layout. Of course you could notice some related issues that I didn't as you're more familiar with plugin. Spilling country name onto 2 lines is not the worst scenario - it is OK for many users. Especially if user types first letters of country name. Probably it may be more tricky if it won't be splitted in few lines and some country names will be overlapped. Thanks for the response and great work on this plugin!

jackocnr:
Thanks for your reply. This is the first time I have heard this use case and I'm currently trying really hard not to add new features unless they would benefit lots of people, so I think for now I will advise you to do a workaround, and if enough people express an interest in this feature then I will reconsider.

For a workaround you could do something like the following (immediately after initialising the plugin):

```js
$(".intl-tel-input .country-list").width($(".intl-tel-input input").outerWidth());
```

vitaliy-patsay:
OK, thanks. Good luck!

PR #432, opened 2016-06-19
Update README.md with Typescript type definitions

fdnhkj:
Added information about intl-tel-input.d.ts file for Typescript.


jackocnr:
Thanks for this. I'm trying to reduce the number of sections in the readme, so in the end I opted to just add it to the links section.

Thanks again - this is a super cool feature!

PR #435, opened 2016-07-06
Added initialCountry="blank" and flagOnEmptyDialCode options

unpollito:
I'd only want to display a flag whenever the user has entered a valid dial code or chosen a flag in the dropdown list. Therefore, I've added two new options:

- initialCountry: "blank": allows the input to be initialized with no country (and thus, no visible flag) selected.
- flagOnEmptyDialCode: if false, does not show a country flag whenever the dial code is deleted after a correct number was introduced. For instance, if I write "+33", a French flag appears; by default, if I then empty the input, the French flag remains; setting this flag to false ensures that the flag disappears.


jackocnr:
Can you explain why you think having the flag default to an empty state would be a better user experience? I have purposefully have avoided this by always setting a default country, and reverting back to that if the user clears the input. I think it's much clearer to the user to see a selected flag, then they immediately understand the purpose of that element, and if it's the wrong flag then they know to change it.

I think leaving the input in an empty state is a confusing user experience. It's not like a select element, where you can display an empty message to the user, it's just a grey rectangle - it's not intuitive at all. I would be very concerned about users just ignoring it, or misunderstanding it somehow. Remember: you should assume users are drunk idiots!

unpollito:
I am currently working on an application which will be used in different countries, so selecting a single default country for them is not acceptable. Furthermore, I can't rely on browser locales to select a flag, since often the country will be wrong (i.e., lots of computers in UK have an en-US locale); nor can I use IPs since the application is run in an intranet and thus users might not even have an Internet connection. And setting an empty default flag doesn't work since this defaults to the first available country (Afghanistan).

On one hand I understand your point regarding an unclear UI, but on the other, in my particular situation I can't ensure that an appropriate flag will be shown, so I'd rather not display any.

jackocnr:
I disagree - I think it provides a better user experience to have the wrong flag selected initially because it makes it clear to the user that it is a country select element. If you just have a grey rectangle there, I think a lot of people will find it confusing, and I think it's likely some people would just ignore it. Also if you don't have a selected country, then it doesn't make sense to have a placeholder number either, which is a standard element of good UX.

And when you say your app will be used in lots of different countries - just to be clear - that's not unique to your situation - that's actually the typical use case for this plugin - that's exactly what it was designed for.

I'm sorry but I don't think this would be a good addition to the plugin right now, especially seeing as no-one else has requested this feature, so I'm going to close this issue for now. If this becomes a popularly requested feature then I will re-consider.

PR #450, opened 2016-07-18
Rename sprite-retina-mustache.scss to sprite.scss.mustache to avoid compass error

kfiroo:
Also changed reference in /grunt/sprite.js


jackocnr:
Thanks!

PR #456, opened 2016-07-28
.

toxaq:


jackocnr:
Wouldn't the `dropdownContainer` option solve your problem?

toxaq:
Not from what I can intimate from the description of it. I'm just using this to add and remove a class from the row container which then can trigger CSS animation. The `dropdownContainer` functionality seems to relate to the closing of the dropdown on scrolling. Using my method I'm can eliminating the scrolling (the scrolling did work fine for me, I just didn't want any scrolling). Hopefully my gif explained the usecase better than I'm doing here.

jackocnr:
Sorry I think the dropdownContainer description was misleading - I have re-phrased it - does that make more sense?

Basically if you just set dropdownContainer to "body" then the dropdown should work fine!

toxaq:
I guess my issue is there was nothing wrong with the dropdown functionality at all. I just don't want my users to have to scroll in this instance. It's UX not a UI problem. I'm not following how changing the scroll container would make the rendering of my form bigger?

jackocnr:
> I'm not following how changing the scroll container would make the rendering of my form bigger?

It wouldn't make your form bigger, it would just mean the dropdown would be able to overflow the form.

> I just don't want my users to have to scroll in this instance

Using the dropdownContainer solution wouldn't have any impact on the scrolling needed. The dropdown would appear in exactly the same place on the screen with either of the proposed solutions. Does that make sense?

I think of the country dropdown as being a prettified `<select>` element, and a select element would just overflow the form.

toxaq:
Yes, I think I'm understanding exactly what the dropdowncontainer does and to me it won't help with the outcome I've achieved. Did you see my gif? I put it behind a details tag so it doesn't trigger anyone with flashing light issues (it's a bit flashy on repeat).

caseyjhol:
Using `dropdownContainer` would provide the exact same functionality as in your GIF, without resizing the modal. This is preferred, as it's poor UX to have the "Save" button moving around on the page.

Issue #54, opened 2014-08-02
Publish to npm


lloydwatkin:
I'm using `browserify` for my project, so it would be great to see this on npm

ConAntonakos:
What's the best way to expose the npm module? I'm simply requiring it and trying to expose it within the compilation process via webpack, but I'm getting an empty object. Any ideas?

lorensr:
@jackocnr I have a similar problem – after switching from Bower to npm, I've tried including `'intl-tel-input'` and `'intl-tel-input/build/js/intlTelInput.js'`, but with both I get the error `TypeError: $(...).intlTelInput is not a function`. Should just including get your module to add itself to jquery, or is there something I need to do? Perhaps it's not being loaded because there's no `main` in your `package.json`? 

A couple other examples are [sweetalert](https://github.com/t4t5/sweetalert/blob/master/package.json), which after including, a global is available, and [node-waves](https://github.com/fians/Waves/blob/master/package.json), which I have to assign to a variable myself.

jackocnr:
So I haven't done this for a while, but I can tell you how we were doing it in the last project I was working on a few months ago: we used npm to install intl-tel-input (instead of bower), and we used browserify to build our JS app, and the way we did it was to have a global jquery, and then in the view that wanted to use this plugin, I just did `require("intl-tel-input");` at the top (I didn't assign it to a variable), and then to use it I just did `this.ui.phone.intlTelInput();` - this is a Marionette.js ItemView, so this.ui.phone is the jQuery object for the input.

pjebs:
Is the above instructions still valid?

I have jquery in global namespace.
I am using browserify.
I have done  `npm install intl-tel-input --save`
I have added `require("intl-tel-input");` at the top of file.

I tried to use: `$(this.refs.input).intlTelInput();`, where `this.refs.input` refers to actual input box in DOM.

**The error I get is: `TypeError: $(...).intlTelInput is not a function`**

pjebs:
I tried it with `$(this.refs.input).intlTelInput();` too. Same result.

jackocnr:
Nothing has changed so that really should still work. Unfortunately I personally no longer work with Browserify so I can't confirm that, but maybe someone else can?

pjebs:
I gave up. Just using the html script tag.

andreiho:
Using a shim fixed it for me:

```
"browser": {
  "jquery": "./node_modules/jquery/dist/jquery.js"
},
[...]

PR #470, opened 2016-08-30
Remove jquery from dependencies

piuccio:
Fixes some of the comments in jackocnr/intl-tel-input#54

I'm trying to use code like this

```js
require(['jquery', 'intl-tel-input'], function ($) {
   $('.phone').intlTelInput();
});
```

But I get the error

```
TypeError: $(...).intlTelInput is not a function
```

Took me a while to figure out, but it turns out that webpack includes jQuery twice in the bundle.

```js
webpackJsonp([1],{
   2: function(module, exports, __webpack_require__) {
       // jQuery imported by my script
   },
   3: function(module, exports, __webpack_require__) {
      /*
	 * International Telephone Input v9.0.8
	 * https://github.com/jackocnr/intl-tel-input.git
	 * Licensed under the MIT license
	 */
	// wrap in UMD - see https://github.com/umdjs/umd/blob/master/jqueryPluginCommonjs.js
	(function(factory) {
	    if (true) {
	        !(__WEBPACK_AMD_DEFINE_ARRAY__ = [ __webpack_require__(4) ], __WEBPACK_AMD_DEFINE_RESULT__ = function($) {
	            factory($, window, document);
	        }.apply(exports, __WEBPACK_AMD_DEFINE_ARRAY__), __WEBPACK_AMD_DEFINE_RESULT__ !== undefined && (module.exports = __WEBPACK_AMD_DEFINE_RESULT__));
	    } else if (typeof module === "object" && module.exports) {
	        module.exports = factory(require("jquery"), window, document);
	    } else {
	        factory(jQuery, window, document);
	    }
	})(function($, window, document, undefined) {
             // intl-tel-input
       })
   },
   4: function(module, exports, __webpack_require__) {
      // jQuery requireb by intl-tel-input
   }
})
```

Because jQuery is marked as an explicit dependency of `intl-tel-input`, webpack includes both `node_modules/jquery` and `node_modules/intl-tel-input/node_modules/jquery` in the bundle.

I've tried [DedupePlugin](https://webpack.github.io/docs/list-of-plugins.html#dedupeplugin) but that didn't work.

The simplest fix is to set `jquery` as a peer dependency (which is exactly what it is).

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   ],
   "license": "MIT",
   "author": "Jack O'Connor (http://jackocnr.com)",
-  "dependencies": {
+  "peerDependencies": {
     "jquery": ">=1.7"
   },
   "devDependencies": {


jackocnr:
Thanks. This makes sense. Reference: https://nodejs.org/en/blog/npm/peer-dependencies/

jackocnr:
Released in v9.0.9

piuccio:
Thanks!

PR #476, opened 2016-09-12
Change autoPlaceholder options to accept off, polite, or aggressive.

jtwalters:
- Allows overwriting the placeholder (see #412)
- Keeps backward compatibility of Boolean false/true from previous versions.


jackocnr:
This is great thank you. Some minor feedback:

1. Please can we add some extra parens to the new condition so we don't have to remember which takes precedence between && and ||
2. Why did you remove the quotes around the "class" property in `_generateMarkup`? I thought that you had to put quotes around reserved words like that?

Thanks!

jtwalters:
I added extra parens in the source code for clarity. Please note that uglify then removes the extra parens and is also removing the quotation marks around the `class` object properties. I ran `grunt build` and that was the result.

jackocnr:
Interesting. What version of grunt-contrib-uglify do you have installed? (I'm using v0.6.0 and it doesn't strip the quotes).

jtwalters:
Ok, so it wasn't `grunt-contrib-uglify` but more likely `uglify-js2`. I found out you can preserve the quotes by specifying `screwIE8: false`. Maybe only IE8 has a problem with it?

jackocnr:
Haha so funny story - as part of another PR, I upgraded all of our libs, and so now the class quotes have been stripped in master (which I've decided is fine), and so now this PR is re-adding them! I'm sorry - this has been a bit of a journey...

If you could rebase to master (and do an `npm install` to get the new versions of everything), and remove that screwIE8 stuff, then this should be good to merge 👍

jtwalters:
Actually it breaks IE8 and think you should preserve those quotes. I know
that browser is ancient but if that's the only breaking point I think it's
worth keeping. Btw, I did rebase and npm install. This PR is built with the
latest dependencies.

jackocnr:
Yeah I guess you're right. I mean I really want to drop IE8 support ASAP, but maybe it's not time just yet.

jackocnr:
Released in v9.1.0, thanks!

PR #482, opened 2016-09-23
Force GET type ajax call for Utils script

cadams93:
jQuery's ajax method is pre-loaded with settings from previous calls (in my case a POST request) causing a POST request to the js file which returned a 405 Method Not Allowed. Fix by forcing GET type request to the resource ```

 ``` ```

 ```


jackocnr:
Sounds good, thanks.

jackocnr:
Released in v9.0.13

PR #487, opened 2016-09-25
Make defaults object accessible from outside

samosad:


jackocnr:
Would you be able to add a test for this?

samosad:
done,
after I added a test, I found a problem - if defaults object is accessible from outside than its keys shouldn't be replaced with one-symbol names. So I removed some code from `replace` task to make tests pass again.

src/spec/tests/static/defaults.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+describe("defaults: change default plugin options", function() {
+
+	var defaults = $.extend({}, $.fn.intlTelInput.defaults);
```
jackocnr: Did you mean to use this somewhere? I guess in the afterEach, to reset the defaults again?

samosad: yes, my bad) fixed

jackocnr:
Released in v9.2.0 - thanks again!

PR #498, opened 2016-10-22
Remove jquery functions isNumeric and extend

alaycock:
I'm not sure if the placement of `_extend` is appropriate, but I'm not sure where else would be best to match your library's convention. Let me know if it's in the wrong spot, and I can move it.

`$.extend(...)` is also used [here](https://github.com/jackocnr/intl-tel-input/blob/00f6ae36c081d5a01facbd12dc263772196f06dd/src/spec/tests/static/defaults.js#L5), although it doesn't do anything, so that can be removed at any time.


tmorehouse:
`$.extend(...)` in this case could be replaced with `Object.assign`, as it appears the defaults are shallow:

`var defaults = Object.assign{{}, intlTelInput.defaults)`

Supported in modern browsers but might need a polyfill for older browsers.

jackocnr:
@alaycock thank you very much for having a crack at this! I agree with @tmorehouse that we should use `Object.assign` here. FYI I have added a note at the top of the [Removing jQuery](https://github.com/jackocnr/intl-tel-input/wiki/Removing-jQuery-Dependency) wiki page to say that I think we should use native methods as much as possible, even if they don't have universal support.

Also I'm so sorry I haven't merged in master for ages, so this PR is full of noise! Let me merge in master now, and then please can you do a new clean PR? Apologies for that.

alaycock:
Sounds good, I'll open a new PR when I have a chance.

tmorehouse:
For reference on `Object.assign`:

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

Note that no versions of Internet Explorer support `Object.assign`, although Edge does.

A simple polyfill (from MDN):
~~~~~~~~
if (typeof Object.assign != 'function') {
  (function () {
    Object.assign = function (target) {
      'use strict';
      // We must check against these specific cases.
      if (target === undefined || target === null) {
        throw new TypeError('Cannot convert undefined or null to object');
      }

      var output = Object(target);
      for (var index = 1; index < arguments.length; index++) {
        var source = arguments[index];
        if (source !== undefined && source !== null) {
          for (var nextKey in source) {
            if (source.hasOwnProperty(nextKey)) {
              output[nextKey] = source[nextKey];
            }
          }
        }
      }
      return output;
    };
  })();
}
~~~~~~~~

jackocnr:
Thanks for the info @tmorehouse.

@alaycock re: isNumeric - I've since realised we don't need a full (generalised) isNumeric function because we're only ever checking a single character, so I think instead we can just do `!isNaN(parseInt(c, 10))` (from [this stackoverflow answer](http://stackoverflow.com/a/8935649/217866)). Also seeing as there's only one case of this, instead of creating a private method, let's just put it inline, with a comment like "check if character is numeric - see http://stackoverflow.com/a/8935649/217866"

PR #502, opened 2016-10-27
Fix - Margin issue in some WP themes

shivapoudel:
In most of the WordPress theme content area list items has some margin styles and to prevent the highlighted gap we have to reset the margin for this selector.

capture

### DIFF
--- a/src/css/intlTelInput.scss
+++ b/src/css/intlTelInput.scss
@@ -191,6 +191,7 @@ $mobilePopupMargin: 30px;
 
     // each country item in dropdown (we must have separate class to differentiate from dividers)
     .country {
+      margin: 0;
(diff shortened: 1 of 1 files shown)


jackocnr:
Hi @shivapoudel, thanks for submitting this. There are a few issues I'd like to address before merging this.

- The divider in the country list (below the preferred countries) is also a list item, so your proposed fix (targeting `.country`) will not work for that (i.e. I think there will still be a gap to the left of the divider) - is that right? Instead I would propose targeting `.country-list li`
- I think normally these things only add left margin, so I think in this rule, we only need to say `margin-left: 0;` - would that solve your problem?
- Finally, I don't think you need to add the rule twice. I think the first one is enough.

So I propose you change it to something like this: `.country-list li {margin-left: 0;}`. What do you think?

jackocnr:
Closing due to inactivity.

shivapoudel:
@jackocnr can you please commit the changeset. I am on a trip without my pc :)

PR #522, opened 2016-12-17
add support for meteor

ziedmahdi:
I added support for meteor. The package is working as expected in a meteor app. The only problem is that I'm not familiar with grunt, so the grunt file may need some modification.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -3,6 +3,28 @@ module.exports = function(grunt) {
   // load all tasks from package.json
   require('load-grunt-config')(grunt);
   require('time-grunt')(grunt);
+  
+  grunt.initConfig({
+    exec: {
+      'meteor-init': {
+        command: [
+          // Make sure Meteor is installed, per https://meteor.com/install.
+          // The curl'ed script is safe; takes 2 minutes to read source & check.
+          'type meteor >/dev/null 2>&1 || { curl https://install.meteor.com/ | sh; }',
+          // Meteor expects package.js to be in the root directory of
+          // the checkout, so copy it there temporarily
+          'cp meteor/package.js .'
+        ].join(';')
+      },
+      'meteor-cleanup': {
+        // remove build files and package.js
+        command: 'rm -rf .build.* versions.json package.js'
+      },
+      'meteor-publish': {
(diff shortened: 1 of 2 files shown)


jackocnr:
Thanks for putting this together. As a reference for myself and others, here's a link to [Bootstrap's package.js](https://github.com/twbs/bootstrap/blob/master/package.js).

Some feedback: in package.js, why do you include 'ecmascript' in the `api.use` call? (I noticed that Bootstrap don't do that). Also can you tell me why you added `bare: true` at the end?

Finally, can you explain all of your changes in the Gruntfile? We use [grunt-bump](https://github.com/vojtajina/grunt-bump) to update version numbers - perhaps that will work with this as well? You should just need to add it to the list of files in grunt/bump.js.

ziedmahdi:
`api.use`is for declaring dependencies.

`ecmascript` is to enable ES6 syntax, as from version 1.2.1 meteor support ES6. As we are not exporting anything, we can remove the ecmascript dependency.

`bare: true` to tell meteor not to wrap those files in a closure. I got some problems before with certains html template and jQuery plugins and using this options solved the problem.

I have used the meteor package in a meteor app (version 1.2.1) and it worked perfectly. Great work by the way.

I'm not familiar with Grunt, what I wanted to do is to add a task for publishing the package to atmosphere (it's like publishing nodejs module to npm)

jackocnr:
Ok I found https://github.com/MeteorPackaging/grunt-gulp-meteor which helps me understand your approach.

Changes required to get this merged: 

1) Remove the "ecmascript" part
2) Move the grunt exec rules into the file grunt/exec.js (we're using the load-grunt-config plugin, which allows you to put these rules in separate files to keep things neat)
3) Get rid of the dupe "meteor" grunt task (as it just runs the same tasks as "meteor-publish")
4) Add "meteor-publish" to the end of the task array on the line that starts `grunt.registerTask('version'`, so it actually gets called!
5) Find a way to auto-update the version number in package.js - hopefully this is possible with the grunt-bump plugin we're currently using, else you will need to find another way.

I'm closing this for now - please submit a new PR when you're ready. Cheers.

PR #524, opened 2016-12-28
Adding protocol jquery URL

jugutier:
Otherwise it can't find jQuery.

### DIFF
--- a/demo.html
+++ b/demo.html
@@ -16,7 +16,7 @@
   </form>
 
   <!-- Load jQuery from CDN so can run demo immediately -->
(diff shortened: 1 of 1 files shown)


jackocnr:
Google used to recommend you leave the protocol unspecified like that, so it uses the protocol of the page you're on, but now it seems they recommend you always use HTTPS, so thanks!

jugutier:
@jackocnr I think that still happens, it's just that if I downloaded the component in a zip file or cloned the repo and double click on the file (which was my case) the base protocol will be file:// and that means jquery would have to be locally available (i.e /etc/hosts/ would have to have a rule to resolve ajax.googleapis.com) Please let me know if this line of thought is wrong.

PD there is another occurrence of this on the README.md that might be worth correcting also.

PR #526, opened 2016-12-30
recommend HTTPS for $.get in geoIpLookup sample

ghiculescu:
without this fix, this sample code will fail when run from a HTTPS page


jackocnr:
Unfortunately ipinfo requires a paid account to use over HTTPS. Instead I have updated the readme to mention that.

PR #534, opened 2017-01-09
Add getPhoneObject to return code and number in object

Massad:


jackocnr:
I'm afraid I don't see a common use case here. You can just use `getSelectedCountryData` and `getNumber` (and specify you want national format, if that's what you want).

PR #557, opened 2017-02-19
Change grunt build to npm run build

tinovyatkin:
It's a pain for a contributor that doesn't have `grunt` installed globally to run build task as `./node_modules/.bin/grunt build`. This PR fixes it to just `npm run build` and allows to change build system in the future without updating docs.


tinovyatkin:
Just a build tools small fix

jackocnr:
Well, it allows you to run the `build` task through npm, but not any of the other tasks e.g. `watch`, or I personally use the `js` task over the `build` task because it's so much faster.

Why can't you just do a `npm install -g grunt`?

tinovyatkin:
Well, I most certainly can install globally anything on my dev machine, but this PR is about making a contributor life a little bit easier and docs consistent (your docs says nothing that you expect globally installed `grunt` or Ruby-based globally installed `sass`).
You already have `test` command in NPM scripts, so, it's not a virgin anyway :)
`watch`, however, is mostly developer preference, I agree that adding it will not harm anybody and may do it easily if you want.
I mean, most of the modern libraries (`eslint` and `babel`, to name a few) recommend local, per project installation, to make project self-containing. It's also good for CI systems, etc - so, doesn't require a potential contributor to install nothing globally for making quick PR to your project - it's a good practice as far I understand today' Javascript world.

jackocnr:
I wanna start by saying thank you - again, I appreciate your intention. And yes I agree that ideally this would be a completely self-contained project, and I would be willing to consider this PR if it achieved that goal completely, but it doesn't yet - as you say we still require a globally installed sass, and evenizer as well. So until we can achieve that goal, I'm not convinced this is the right route to go down. Instead I would favour a simple clarification in the wording in CONTRIBUTING.md.

tinovyatkin:
My next PR gonna be switch to `grunt-sass` to remove dependency on Ruby sass, but I can't do two PR in parallel...

jackocnr:
Cool! Please can we push all commits to the same branch (this PR) until we're ready to merge.

tinovyatkin:
Have no idea where do you use `evenizer`, but it's mentioned in docs, so, I see no problem with it... `node-sass` will allow to switch from mustache template for sprites to a native Js if you want to later...

tinovyatkin:
Not sure what happening with `travis` as `npm test` shows no failing test locally...

jackocnr:
Please don't remove the build dir, as [cdnjs](https://cdnjs.com/libraries/intl-tel-input) relies on that, and also it's nice for people to be able to download the plugin and get the demo up and running straight away.

tinovyatkin:
Oops, sorry

tinovyatkin:
It was kinda strange to see `build` dir tracked by `git`

tinovyatkin:
Just read `cdnjs` docs and it seems to pull files from `npms`, so, I think the correct way is not to track `build` dir (that's rare to track autogenerated files with git), but instead create `prepublish` task at `packages.json` and add `build` dir into `files` section for `package.json`. Did you consider that way?

tinovyatkin:
:tada: no globally installed dependencies. Still need to make `travis` run build just once (it currently runs both, `postinstall` and `prepublish` tasks for `npm install`)

tinovyatkin:
I should say that the build process for `utils.js` is crazy, but from my point of view should either be included into project scripts or outsourced for something like https://github.com/seegno/google-libphonenumber

jackocnr:
Haha yup the build process for utils.js is a PITA - I would be open to a PR to change this!

That project you linked to looks great, but their build is twice the size of ours, so that's not feasible.

tinovyatkin:
@jackocnr it looks pretty done for me, what do you think?

jackocnr:
Thanks for this!

Please can you undo your changes in .gitignore and re-add the build files.

jackocnr:
Oh I see, you're saying we no longer need the build files. Hmm I need to think about this.

jackocnr:
Both Bootstrap and jQuery have their built files in their repos...

I think it's required for when people are using the plugin in their projects (rather than cloning this project directly to contribute to it). People can just add it to their dependencies in their package.json and `npm install` and then import the module.

tinovyatkin:
They have build files tracked by `git` (wow, that even sounds rare) due to `bower` (that's the strange piece of distribution). Will get back `build` files now.

tinovyatkin:
All done! I'm actually have another, functional PR ready (to fix options parameters mutation), just was thinking this one will be small and fast to setup good base for contribution :)

jackocnr:
Great thanks! However, there's still a couple of things to clear up here before we merge:

- In a couple of places (.gitignore and package.json) you have removed the new line char at the end of the file - please undo these changes, as [it is bad practice](http://stackoverflow.com/questions/5813311/no-newline-at-end-of-file)
- This PR is just about build commands, right? So why do we have [changes in the built JS, CSS and images](https://github.com/jackocnr/intl-tel-input/pull/557/files)? (Note: regarding the image changes, I sometimes find git randomly detects changes when I do a build - but AFAICT nothing has changed - probably just a timestamp in the metadata etc, so I don't commit the changes, to keep the commits clean and meaningful). Please can we remove all changes to files in the build directory for this PR?

PR #571, opened 2017-03-27
refactor _getDialCode

tinovyatkin:
This PR fixes two problems:
1.  _getDialCode will not throw on null number (I’m accidentally got to this error while using `setNumber` with undefined value from localStorage)

2. removed JQuery method use


build/js/intlTelInput.js
```diff
@@ -886,8 +882,7 @@
                     }
                     // a lot of numbers will have a space separating the dial code and the main number, and some NANP numbers will have a hyphen e.g. +1 684-733-1234 - in both cases we want to get rid of it
                     // NOTE: don't just trim all non-numerics as may want to preserve an open parenthesis etc
-                    var start = number[dialCode.length] === " " || number[dialCode.length] === "-" ? dialCode.length + 1 : dialCode.length;
-                    number = number.substr(start);
+                    number = number.replace(dialCode, "").replace(/^[\s-]*/, "");
```
tinovyatkin: This is just a bonus, it will remove one OR MORE spaces after dial code...

jackocnr:
Thanks a lot for this Tino 💐. I'm very busy at the moment, but will take a proper look as soon as I can.

src/js/intlTelInput.js
```diff
@@ -1063,7 +1058,8 @@ Plugin.prototype = {
   _beforeSetNumber: function(number) {
     if (this.options.separateDialCode) {
       var dialCode = this._getDialCode(number);
-      if (dialCode) {
+      // _getDialCode returns empty string if there is no dialCode
+      if (dialCode && dialCode !== "") {
```
jackocnr: FYI this second check is redundant - `if (dialCode) {` will already be falsy if dialCode is an empty string - most likely I use this technique in a few places throughout the code.

tinovyatkin: Oops, sure, I was thinking in strict equality... Will remove it now for sake of beauty!

jackocnr: Beauty 😍

jackocnr:
Ok so thanks again for this PR - I've just spent some time going over it and it contains some interesting ideas. And the code works 👍 - all the tests pass! But I have to say I find it much less readable/intuitive than the current implementation.

You mentioned the reason for this PR is to fix 2 problems: throwing an error when passing null, and removing jQuery's $.isNumeric, but those two problems could be solved with simple 1-line changes. The error would be solved by modifying that first if condition to start with `if (number &&`. And $.isNumeric can be replaced by a simple regex `/\d/.test(c)`.

So what's the reason for the big refactor?

tinovyatkin:
@jackocnr when a code have a test coverage then I'm always use the opportunity to improve/refactor that code while fixing a problem.
In your original code you starting from shorter dial code and progressing to a biggest match, I'm doing the reverse using RegExp. It's a matter of personal preference and small benchmarks, that a not critical, but you certainly, as a project owner, may do it your preferred way.

jackocnr:
Haha. Well I do like your thinking - starting with the 4 numeric chars and working backwards - it has the potential to be more efficient. But in this implementation, with the splitting and joining of arrays, and the working backwards though the `dialCode` string using regex replacements... as I said, I just find it difficult to read and understand.

That said, if we did some benchmarking and found this implementation to be faster, then I think we should go with it! And btw if you're interested in looking into this, I think there are a couple of optimisations we could make: firstly, in the `exec` call, we could put parens around each of the 4 digits, so we get them separately in the results array `rawDial`, and then we could call `rawDial.shift()` to magically get the `digits` array without having to call `replace` and `split`! And secondly, we could inline the regex in the `dialCode.replace` statement, and remove the trailing `\D` as I don't think it's ever needed.

tinovyatkin:
@jackocnr You are right, there are always space for improvements. Did some changes as you suggested.

jackocnr:
Bad news I'm afraid. The new implementation is slower: https://jsperf.com/getdialcode/1

Let me know if you have any feedback/thoughts on that test, or if you can think of any ways to improve the perf of the new implementation!

jackocnr:
I'm going to close this for now. As I say, let me know if you can find a way to make this faster than the current implementation. Alternatively, it would be great to get a PR that fixes the issues mentioned here with the 1-liners I mentioned.

PR #573, opened 2017-03-27
isGoodBrowser seems to be broken

tinovyatkin:
You are testing property on `element` (that can be a string selector), instead of the actual `input` element?

### DIFF
--- a/src/js/intlTelInput.js
+++ b/src/js/intlTelInput.js
@@ -62,7 +62,7 @@ function Plugin(element, options) {
   this.ns = "." + pluginName + (id++);
 
   // Chrome, FF, Safari, IE9+
-  this.isGoodBrowser = Boolean(element.setSelectionRange);
+  this.isGoodBrowser = Boolean(this.telInput[0].setSelectionRange);
 
   this.hadInitialPlaceholder = Boolean($(element).attr("placeholder"));
 }


jackocnr:
Ha, you had me confused for a second there - it's been a while since I looked at this code, but no it's currently working as expected. So when you init the plugin, you're doing that in the context of an existing jQuery selection e.g. `$('#phone').intlTelInput()` - it doesn't get passed the string, it gets passed the jQuery object(s) resulting from `$('#phone')`. Then in the plugin code [here](https://github.com/jackocnr/intl-tel-input/blob/master/src/js/intlTelInput.js#L1248) we iterate over that selection, accessing the raw DOM elements, and passing those into the `Plugin` constructor. Does that make sense?

tinovyatkin:
I must confess I never wrote a JQuery plugin...
So, if `element` is already a JQuery element, why you do below:
`$(element).attr("placeholder")` instead of `element.attr(..)`? Just trying to understand code before touching `$.extend`...

jackocnr:
`element` is not a jQuery element, it is a raw DOM element, which is why we do `telInput = $(element)` at the top of that function. One optimisation we could do is to use `this.telInput.attr("placeholder")` instead of converting `element` to a jQuery object again.

tinovyatkin:
I'm actually was looking for a way to save reference to DOM element and reuse it elsewhere while replacing JQuery method, so, we should do something like `this.element = element`?

tinovyatkin:
I think targetting leaving JQuery it's better to do just `!!element.placeholder`

jackocnr:
Sure we could add a property for the dom element, though at this point I think we should be explicit and call it something like `this.domElement` to avoid confusion.

And yes I agree with just using `element.placeholder`, but I'd rather stick to readable code like `Boolean()` instead of `!!` which IMO is an byte saving optimisation which belongs in the build process.

Speaking of the build process, I would also be open to a PR that introduced a Babel build step (targeting IE11, which is all we will have to support come April 4th), to allow us to write in ES6!

tinovyatkin:
I was afraid to offer you rollup + Babel, but will do it with a pleasure! I'm strugling without ES6 here!

PR #605, opened 2017-06-23
add event listener for close:countrydropdown

willashley23:
I added an event listener in `intelTelInput.js` for closing the country code dropdown. I added tests in the core test directory, basing it off of the tests for the `countrychange` event. I also added a helper method that is used inside the test spec. 

The idea is that you should now be able to do this:

```
$phoneInput.on('close:countrydropdown', function() {
   // do stuff.
});
```

If this checks out ok, I'll gladly do the same for opening the country dropdown, but since this is my first time contributing, I want to make sure I get things right.


willashley23:
@jackocnr Does this check out?

src/js/intlTelInput.js
```diff
@@ -920,6 +921,8 @@ Plugin.prototype = {
 
   // close the dropdown and unbind any listeners
   _closeDropdown: function() {
+
```
jackocnr: Please can we get rid of these extra new lines? (here, and in `_showDropdown`)

src/js/intlTelInput.js
```diff
@@ -938,6 +941,9 @@ Plugin.prototype = {
       }
       this.dropdown.detach();
     }
+
+    this._triggerCloseCountryDropdown();
```
jackocnr: No need to create a new function for this just yet, if it's only called in one place, and it's just a one-liner anyway.

src/spec/helpers/helpers.js
```diff
@@ -82,6 +82,10 @@ var selectFlag = function(countryCode, i) {
   getListElement(i).find("li[data-country-code='" + countryCode + "']").click();
 };
 
+var openCountryDropDown = function() {
+    $(".selected-flag").click();
```
jackocnr: Please use `getSelectedFlagContainer().click()` here.

src/spec/tests/core/closecountrydropdownEvent.js
```diff
@@ -0,0 +1,41 @@
...
+      input.intlTelInput();
+    });
+
+    it("does not trigger the event", function() {
+      expect(spy).not.toHaveBeenTriggered();
+    });
+
+    it("clicking outside the dropdown to close it triggers the event", function() {
+      openCountryDropDown();
+      $("<h1>").click();
```
jackocnr: I'm not sure if this will work, if the element is not in the DOM? Instead, let's inject the input into the DOM in the beforeEach (see example [here](https://github.com/jackocnr/intl-tel-input/blob/master/src/spec/tests/core/dropdownShortcuts.js#L7), and remove it in the afterEach) and then use `input.click()` - does that work?

src/spec/tests/core/closecountrydropdownEvent.js
```diff
@@ -0,0 +1,41 @@
...
+      expect(spy).not.toHaveBeenTriggered();
```
jackocnr: Shouldn't this be `expect(spy).toHaveBeenTriggered();`?

build/js/intlTelInput.js
```diff
@@ -483,6 +484,7 @@
         },
         // show the dropdown
         _showDropdown: function() {
+            //findme
```
jackocnr: Please re-build the JS to get rid of these :)

src/js/data.js
```diff
@@ -929,7 +929,7 @@ var allCountries = [
     "250"
   ],
   [
-    "Saint Barthélemy (Saint-Barthélemy)",
+    "Saint Barthélemy",
```
jackocnr: 👍

jackocnr:
@willashley23 this is great thank you!

I've added a few little cleanup requests, but otherwise this is brilliant, thank you so much :)

Also if you want to go ahead and add the open event as well, that would be great. Or we can handle that separately if you'd prefer.

willashley23:
@jackocnr I made many of the changes you suggested. I tried to remove the one-line function as you suggested, but every time I run the build script, it replaces it with what I had before.

willashley23:
@jackocnr Also, I can't figure out why that one test is failing.
I can't seem to run the test script with it reflecting my changes, either. It only runs the build version, and I was going to try and move the call to trigger the `close:countrydropdown` event to higher up in the function. 

It could also have something to do with `_closeDropdown` not getting called. It seems to work fine in all other cases. Any ideas?

jackocnr:
@willashley23 re: removing the one-line function, and it getting replaced when you run the build script - my guess would be you're updating it in the build/js/intlTelInput.js, but instead you need to update it in src/js/intlTelInput.js.

willashley23:
@jackocnr Yep, that was it. When I typed in the name of the file in sublime it was giving me the build version first. I removed the extra whitespaces in the src/js one and removed the one-line function, though this seems to have created a merge conflict with the build file.

jackocnr:
Don't worry about the build file - it's because I just updated the libphonenumber build, which incremented the version number in the build file. I can sort this. I'll try to merge this on Saturday if I have time.

jackocnr:
@willashley23 released in v11.1.0 - great work, thank you so much 🎉

PR #611, opened 2017-07-10
Added allowTabOnFlag option to enable/disable tab/focus to flag area

Cactusbone:
Added allowTabOnFlag option to enable/disable tab/focus to flag area

Useful to allow direct access to phone input on tab, since typing the international prefix already changes flag

### DIFF
--- a/src/js/intlTelInput.js
+++ b/src/js/intlTelInput.js
@@ -4,6 +4,8 @@ var pluginName = "intlTelInput",
   defaults = {
     // whether or not to allow the dropdown
     allowDropdown: true,
+    // should tab key be able to toggle to flag
+    allowTabOnFlag: true,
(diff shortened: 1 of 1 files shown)


jackocnr:
Thanks for the idea, but I'm trying really hard to keep this plugin as light as possible, and this is the first time I've heard this request. For me, conceptually, dropdowns like this are their own control elements, and deserve their own focus state.

Also by leaving it in, the worst case is the user has to tab twice. By taking it out, it makes it impossible to navigate the dropdown (visually) using the keyboard, which may frustrate some of your users.

Cactusbone:
I understand. My use case is most people visiting my website will have a phone on the default country, and some (like less than 1%) will want to change it, so tabbing directly to the number makes it easier, but yeah it's only another tab away

PR #639, opened 2017-09-22
Add setPlaceholderNumberType public method

cedk:


jackocnr:
Hi there, thanks for putting this together - esp nice to see the documentation and tests in there too. But I definitely would have recommended creating an issue to discuss this before implementing it as I only want to add features that will be useful to a lot of people. Please can you describe your use case?

cedk:
My use case that I have a select to define the type of phone number (ex: Mobile, fixed line etc.), so I want to have the placeholder to adapt to what the user select first.

jackocnr:
Ok fair enough. Unfortunately there's a conflict now because I recently pushed a minor change. If you can rebase and rebuild the JS then I will merge this. Thanks!

jackocnr:
Thank you! Released in v12.1.0.

Issue #644, opened 2017-10-04
Clean country name


juggernaut451:
A function to create a clean country name eg:
Hong Kong (香港) => Hong Kong
Hungary (Magyarország) => Hungary
India (भारत) => India

jackocnr:
Can you explain your use case?

juggernaut451:
Basically, I just want the country name, not the local language name of the country.

jackocnr:
I understand that, but I want to know why. I'm trying to understand if this might be a common requirement or not. I don't want to add weight to the plugin for features that will not be used by many people, when it's easy to just do this in 1 line of JS yourself.

jackocnr:
Closing due to inactivity.

PR #645, opened 2017-10-04
Added a new field  cleanName to the countryData field on the countryC…

hanut:
Pull request against [Issue #644](https://github.com/jackocnr/intl-tel-input/issues/644) . Have added a new field on the country data received in the onCountryChange event callback


jackocnr:
Thanks for putting this together, but first we need to discuss it and agree it's something we need.

PR #660, opened 2017-11-21
Prevent numeric strings from overflowing integer parsing

arkamax2k:
In libphonenumber, number parsing involves calling parseInt(). If the number passed for parsing is larger than 2^53-1, it will be parsed into an integer that is different from the original string representation. The result is that when calling "getNumber" on a long numerical string, one may receive a number that is different from the one entered.


jackocnr:
Thanks for this. I'm afraid I'm really busy with work at the mo, but hope to test this one evening next week.

arkamax2k:
Not a problem, thank you for looking. I think this is somewhat of an edge case, however a possibility of data corruption, especially one that is hard to notice if the field is not long enough, warrants a fix IMO.

jackocnr:
Apologies for the massive delay - I finally found some time! Thanks again.

Issue #667, opened 2017-12-13
Aria and Role attributes for accessibility


MorneL:
This is less of an issue and more of a discussion before I submit a pr. Quite a few aria and role attributes are required to make this WCAG 2.0 compliant.

Expected:
on ul.country-list (aria-expanded: true/false, role: listbox)
on div.intl-tel-input (role: combobox)

There are a few more that I will point out, once I did the dev work.

@jackocnr would this be something you need?

jackocnr:
@MorneL yes I would welcome a PR to add this stuff, thanks :)

PR #674, opened 2018-01-05
Added id's and aria properties

MorneL:
Had to add Id's to each list item for the aria-activedescendant property to work properly, also had to add focus to the highlightListItem for the screen reader to notice which list item is focused.


MorneL:
@jackocnr here are the changes I mentioned in this issue https://github.com/jackocnr/intl-tel-input/issues/667

andre-the-giant:
Hi

Is there any upcoming release including these changes soon ?
I know an expert in Accessibility that could check if it makes this plugin more accessible.. 
Thanks !

jackocnr:
@MorneL I apologise for the delay in reviewing this - I have been very busy. I will get to reviewing this ASAP.

@andre-the-giant thanks for the offer - it would be great if you could get your friend to take a look. You can do this now using [MorneL's branch](https://github.com/MorneL/intl-tel-input/tree/ML-aria-properties).

jackocnr:
@MorneL ok I've finally had a chance to look over this, and it looks great, thank you. A few questions/comments:

- Why the `tabindex` on each list item? My understanding of `tabindex` is that it means the user can hit <kbd>tab</kbd> to focus that element, but that wont work in this case as <kbd>tab</kbd> closes the dropdown.
- Why call `listItem.focus()` in `_highlightListItem`? Seems a bit heavy handed to me, and it means that now all users see a big blue outline in Chrome when browsing the list (even when hovering it with the cursor) which is a bit much. But perhaps it's essential to accessibility in some way?
- Also in this method, we should be able to neaten up the code a bit by chaining calls e.g. `this.countryListItems.removeClass("highlight").attr("aria-selected", "false");`
- With the list item IDs, the country code is unique so we shouldn't need the dial codes in there as well. Instead could we just use the prefix `'iti-item-'`, so for the US it would be `'iti-item-us'`. Does that sound good to you?

Everything else makes sense to me and looks good! Thanks again.

jackocnr:
Hey @MorneL, I'm ready to do a major release - any chance you could push this over the line so we can include it?

jackocnr:
Thanks again @MorneL. I merged your commit and then applied my feedback above in a separate "tweaks" commit, with one change: I moved the "aria-selected" code from the `_highlightListItem` function which is called every time the user hovers a list item, to `_setFlag` which is only called when they actually make a selection.

Issue #405, opened 2016-04-18
Slowly try to move away from jQuery


jackocnr:
I'm making notes on ways we can incrementally move away from jQuery here: https://github.com/jackocnr/intl-tel-input/wiki/Removing-jQuery-Dependency

Feedback appreciated on that wiki page - I'm sure there's a lot of jQuery references in the code that I've missed, and also clever native alternatives!  

I will happily accept pull requests towards this goal (in the branch f-remove-jquery), on the following conditions:
- Each pull request only replaces ONE jQuery feature to keep things clean and simple, and allow discussion
- It doesn't add too much bloat to the code
- They are robust solutions that are unlikely to require maintenance
- The tests pass

Note that in order to make this project feasible, we will only support the latest version of all major browsers, which will enable us to use all standardised native methods (and recommend polyfills for browsers that are not standards compliant).

Thanks for your help!

(Please see wiki page for actionables)

joaocunha:
I want to help with that. Is this up to date?

jackocnr:
I just made a few tweaks to the comment above, and yes I think that wiki page should all still be relevant (even if line numbers are slightly off). As I say: let's stick to removing one jQuery feature per pull request to keep things simple. Thanks!

jackocnr:
@joaocunha FYI I just merged master into that f-remove-jquery branch, so now it's 100% up-to-date!

joaocunha:
@jackocnr awesome! I'll tackle this as soon as I can.

alaycock:
I don't mean to nag, but is there any progress on this? Anything that I could do to help with this?

jackocnr:
Hi @alaycock! You can definitely help. Just implement one of the things on the "Current Actionables" list, and do a PR to the j-remove-jquery branch :)

jackocnr:
UPDATE!!!

Proof of concept here: http://codepen.io/jackocnr/pen/eBbMPY

GLoganDR:
This is great! I've been considering building something like this myself, since most of the angular wrappers for your project are not working properly. Makes it tough to implement when gulp wraps up all of your dependencies during build too :/

jackocnr:
It is not worth the work to remove the jQuery dependency for old versions of browsers - it would be a colossal amount of work to support all of those old versions, with their quirks and their many many bugs! Instead we will only support the latest version of all major browsers with this change - I have updated the wiki page to reflect this.

natcohen:
[...]

PR #676, opened 2018-01-06
removed certain jQuery dependencies

DidierLmn:
I've seen you are trying to move away from jQuery ([#405](https://github.com/jackocnr/intl-tel-input/issues/405)),  and thus I've tried to contribute towards this end.

One of the things I've done is replace `$.ajax` with a `XMLHttpRequest`, however the tests will flag this as issues since they check for `$.ajax.calls`. I've never worked with these tests so I'm uncertain how to adapt/correct these, maybe someone else knows/can help?

Thanks


jackocnr:
Hi there, thanks for putting this together, but I think [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) is the new standard way to do this. There is actually a note about this in the wiki page, which is linked from that issue you mentioned.

sheyooo:
But no IE browser has `fetch` support

jackocnr:
No, that would require a polyfill, which I'm sure a lot of sites are already using.

sheyooo:
I would have loved to help but, I don't know my way around

DidierLmn:
Thanks for the feedback, I overlooked the wiki. I will look it over and try to make another pull request.

Issue #675, opened 2018-01-05
[FEATURE] Bootstrap support


subversivo58:
Bootstrap in its latest development branch is in beta 3 which is a prelude to the release of stable version 4 (announced for the end of this January) and there were significant changes in the [structure of forms and inputs](https://getbootstrap.com/docs/4.0/migration/#forms).

I'm trying to get something close to the figure below using **"addons"** (now suffixed classes for **prepend** and **append**):

![intlinput](https://user-images.githubusercontent.com/6475273/34623026-af60dcba-f236-11e7-963f-df54e4a512fd.PNG)

However this input that by default Bootstrap receives 100% width, but as it is being enveloped in a `div`:

    <div class="intl-tel-input allow-dropdown">

Bootstrap does not recognize (does not apply width) ... to bypass this I tried to use the `w-100` class directly in the `<input>` but the library adds a `padding-right` that does not allow to reach 100% width in addition to not resize this measure using **"break-points"**.

![intlinput-2](https://user-images.githubusercontent.com/6475273/34624480-9af79704-f23c-11e7-8ae7-7ff07dec6c4e.PNG)

> My request

I know this library was not designed to be used specifically with Bootstrap but I wonder if it is possible for some extra class to be added to the `<input>` and correct this? Something like: set `padding-right` using **"break-points"**?

subversivo58:
I tried to use Bootstrap classes and custom classes by following media queries the different **"break-points"** in the quest to regulate the `padding-right` but did not succeed.

In the change of resolution or in different devices visually the field did not occupy all the space or the icon was thrown down.

For the time being the approach that allowed me to make the field visually fluid (following the other elements in all the resolution changes) was to redefine the **css** and add the **"addon"** after starting the library.

The following example using `input-group-sm` (for small input):

```css
.input-group-sm > .intl-tel-input {
  display: flex;
  width: 100%;
}

.input-group-sm > .intl-tel-input > .input-group-append > .input-group-text {
  padding: .25rem .5rem;
  font-size: .875rem;
  line-height: 1.5;
}
```

```javascript

let telInput = $("#phone")

// initialize
telInput.intlTelInput()

// append "addon"
$('.intl-tel-input').append(`
    <div class="input-group-append">
        <span class="input-group-text rounded-0 fa fa-phone"></span>
    </div>
`)
```

Visually satisfying in all changes:

[...]

PR #680, opened 2018-01-17
Add custom class option

subversivo58:
Possibily add custom class to container element.

Reference discussion based from this issue: #675


jackocnr:
Thanks for this! Please could you update the default value to be an empty string, to match the other string based options. And also please can you combine all of these commits into one. Thanks.

subversivo58:
Yes, I think I can do it soon.

chtheis:
I'm looking for exactly this, but I can't find it in the code. Am I missing something?

jackocnr:
I don't think this was ever finished. It would actually be very easy to just do this manually, after you initialise the plugin, either grab the input's parent element, or just select the element with the class `intl-tel-input` and add your class manually.

chtheis:
Am 12.01.2019 um 17:39 schrieb jackocnr:
> I don't think this was ever finished. It would actually be very easy to 
> just do this manually, after you initialise the plugin, either grab the 
> input's parent element, or just select the element with the class 
> |intl-tel-input| and add your class manually.

I can create a patch for it, based on previous work. I don't like to 
manipulate the elements manually ...

I guess the dist/* files are created by some build process and I have to 
patch src/*? The original PR changed the dist/* files which felt strange.

jackocnr:
See the Contributing wiki page for more info: https://github.com/jackocnr/intl-tel-input/blob/master/.github/CONTRIBUTING.md#changes-to-the-plugin

chtheis:
Am 12.01.2019 um 19:06 schrieb jackocnr:
> See the Contributing wiki page for more info: 
> https://github.com/jackocnr/intl-tel-input/blob/master/.github/CONTRIBUTING.md#changes-to-the-plugin

OK, I created the PR. I named the parameter "customParentClass" instead 
of what was in the original PR

npm builds the .js-files without complains.
I did not add a test case: the scripts are beyond me :) And "npm test" 
fails, possibly because it couldn't launch chrome.

PR #687, opened 2018-02-05
Update to jQuery 3.3.1

coliff:
jQuery 1.x and 2.x are not supported and have security vulnerabilities.
This plugin does not support IE8 so it's fine to move to jQuery 3. File size is smaller too so loads faster :-)


jackocnr:
Thanks!

coliff:
cool - will open a PR to update the README code snippet now.

PR #689, opened 2018-02-05
update jQuery version and always load over HTTPS

coliff:
always load over HTTPS makes it faster and safer :-)


jackocnr:
Thanks!

PR #691, opened 2018-02-08
Fix placeholder capping bug

TerrenceLJones:
## Bug: 

intl-tel-input-placeholder-bug

### Observed: 

1. User selects U.S. from the dropdown.
2. User then selects U.K. from the dropdown.
3. User then reselects U.S. from dropdown.

It is observed that the U.S. placeholder is being cut off after reselection. 

### Expected:

The placeholders should be displayed as expected no matter which/when a country is selected.

## Thoughts
This bug essentially occurs when the user has selected a country that has a shorter placeholder and then selects one that has a longer one. The longer placeholder is being cut off in a predictable way. 

It appears that this bug originates in `Plugin#_cap` function in `intlTelInput.js`. 

`_updatePlaceholder` calls `_beforeSetNumber` which then calls `_cap`. `_beforeSetNumber` is called in various scenarios, however the problematic one is when it is called from `_updatePlaceholder.` It seems like the method to control input length is, in this case, cutting off the end of longer placeholders (after previously selecting a shorter one). This seems due to the `maxlength` attribute being out of sync. 

It seems like `_cap` could be bypassed when handling placeholders.

### DIFF
--- a/src/js/intlTelInput.js
+++ b/src/js/intlTelInput.js
@@ -372,7 +372,7 @@ Plugin.prototype = {
   // update hidden input on form submit
   _initHiddenInputListener: function() {
     var that = this;
-    
+
     var form = this.telInput.closest("form");
     if (form.length) {
       form.submit(function() {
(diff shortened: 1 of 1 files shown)


TerrenceLJones:
😢 After trying to get a working example for this PR up on jsbin...and further review of my own code...the `maxlength` in my case is being modified by another library, causing issues with the above code. Sorry for the bother. Here is an working example for reference: http://output.jsbin.com/xekecisava

TerrenceLJones:
😢  While trying to get a working example for this PR up on JS Bin and further review of my own code, I realized that this issue I am having is with another library. The [jQuery-Mask-Plugin](https://github.com/igorescobar/jQuery-Mask-Plugin) is adjusting the `maxlength` in some way causing the above issue. Sorry for bothering with this PR. Here is a working example for reference: http://output.jsbin.com/xekecisava.

PR #697, opened 2018-02-19
Hidden Input Moved After Main Field

hamidyousefi:
In some frameworks like Yii2 and other similar systems, its more efficient to create two field with same name for sending the values we want to post, and because, most of the time we just need to save phone numbers with international form and ignoring the localized form of numbers, if we insert hidden input after main field, its make much more easier way for using this plugin. and the place of hidden input is not important in other systems, specially with different field's name.


jackocnr:
Thanks for submitting this, and apologies for the delay in getting to it.

While personally I would try to avoid having inputs with the same name (it seems like bad practice to me - non-standard and potentially confusing), in this case I see no reason not to merge this change.

Thanks again!

Issue #703, opened 2018-03-05
Countries are not properly sorted with "onlyCountries" option


elonmir:
### Steps to reproduce
1.  Set array in configuration for the option "onlyCountries"
2.  Set array for the option "preferredCountries"

### Expected behaviour
Countries should be sorted alphabetically

### Actual behaviour
Countries are ordered by the order the countries get configured in the options

### Initialisation options
```javascript
{
'onlyCountries': ['BE','BG','DK','DE','EE','FI','FR','GR','IE','IT','HR','LV','LT','LU','MT','NL','AT','PL','PT','RO','SE','CH','SK','SI','ES','CZ','HU','GB','CY'],
'preferredCountries': ['DE', 'AT', 'CH']
}
```

### Before fix screenshot
![image](https://user-images.githubusercontent.com/1140767/36976492-26dd2080-207e-11e8-98ce-d7e7025f6657.png)

### After fix screenshot
![image](https://user-images.githubusercontent.com/1140767/36976528-4230d868-207e-11e8-8240-d7565fa80178.png)

jackocnr:
Yes I suppose this makes sense. Thanks for raising the issue, and submitting the PR. I don't have time to review it right now I'm afraid, but I will try and get to it soon.

elonmir:
You need further support or input on my end?

jackocnr:
Not right now, thank you. I will review the PR when I have time.

elonmir:
Didn't mean to pressure you, just wanted to make sure ;) Take care!

jackocnr:
Merged your PR. Will be included in next release. Cheers.

zeeshanjan82:
As I checked if I am only setting `onlyCountries` as I would like to limit to some specific countries and not have sorting enabled on these but I noticed that sorting is always enabled even if I don't want it in my case I would like to put the countries in a specific order in `onlyCountries` and they should appear in that fashion. Is that possible?

elonmir:
Speaking not on the behalf of the repo owner, I would suggest not necroing this old issue and make a new one. Have you considered using the `preferredCountries': ['DE', 'AT', 'CH']` option?

LieeN:
@elonmir @jackocnr Asking more or less the same question as @zeeshanjan82, is it possible to disable sorting for onlyCountries as I don't want to use preferredCountries option? If I would use a combination of both, the field would show the same country twice (my list of countries is very small, 1-3 countries) which would look quite weird.

[...]

PR #704, opened 2018-03-05
Added sort method for country name

elonmir:
Should resolve issue #703


elonmir:
I just refactored the order of the calls, because with my translation integration the sort order wouldn't have worked.

jackocnr:
Thanks for this.

PR #705, opened 2018-03-05
Feature, translation for countries

elonmir:
Adds the possibility to add custom translations for the countries, if you are not happy with the provided ones. You don't have to run a custom build, especially useful if you just want to use a certain set of countries.


elonmir:
If you wanna push the feature I am gonna add the tests, had to implement the function for a local project.

jackocnr:
Hi there, thanks for submitting this and apologies in the delay in responding.

So up until now I have been against this approach because it just feels wrong to make the user download the main plugin script (including all the names in English), and also download the localized names as part of your custom JavaScript. But until we implement a better solution, I'm happy to go with this approach.

A couple of comments:

- What do you think about defaulting the option to `null`, and then in the code, we can check if the option has been set before iterating through all the countries e.g. `if (this.options.localizedCountries) this._translateCountriesByLocale();`.
- In the readme/comments can we ask for the country codes to be in lower case, to be consistent with the rest of the project (but also call `toLowerCase` on them in the code to be safe).

Otherwise I think we're good to merge! Thanks again.

elonmir:
I'm gonna take a look this week and revise the code.

jackocnr:
Thanks @elonmir, I look forward to getting this merged :)

elonmir:
So take a look, if a missed something.

jackocnr:
Looks good thanks. I'll try and get this merged ASAP.

jackocnr:
Merged and released in v12.4.0. Thanks again.

Issue #719, opened 2018-03-29
utilsScriptDeferred is resolved too early for instances after the first


molnarm:
### Steps to reproduce
1.  Create multiple instances of the phone input, with `utilsScript` specified

### Expected behaviour
Both deferreds returned from creation should be resolved only when the AJAX call to load the script finished.

### Actual behaviour
The deferred for the first instance is resolved only after the utils script is loaded, but the deferred of the second one gets resolved immediately.

### Initialisation options
Set `utilsScript` to a valid value.

### Explanation

This seems to be a regression of #303 , the description there is correct again: in `loadUtils`, there's a check for whether the script loading has _started_, and if that's true, any subsequent calls will get resolved immediately, instead of waiting for completion.

A quick workaround for only this scenario is to _not_ pass the `utilsScriptDeferreds` of instances when calling `loadUtils` from `_initRequests`, because these will be resolved anyway when `handleUtils` is called, but this is not a complete fix (manual calls to `loadUtils` would still work incorrectly).

If you accept pull requests, I created a test for this (failing now, passing with the workaround) [here](https://github.com/jackocnr/intl-tel-input/compare/master...molnarm:concurrency).

jackocnr:
Thanks for reporting this, and writing the test. Yes I welcome pull requests, tho sadly I haven't had enough time to review them in a while. I promise I will get around to it at some point tho!

sentiasa:
```
$.get( "https://cdnjs.cloudflare.com/ajax/libs/intl-tel-input/12.1.12/js/utils.js" ).done(function(data) {
            console.log(data);
       $("#number").intlTelInput({
                   initialCountry: "us",
                   utilsScript: data
        }); 
});
```
Is this the right way?

jackocnr:
Fixed by https://github.com/jackocnr/intl-tel-input/commit/295aa0d468b1d7cfcfb184b1d24fe298a37ac6cd

Let me know if you have any other issues tho.

PR #721, opened 2018-04-03
Handle multiple instances with utilsScript

molnarm:
This branch has a test and a possible fix for #719.

I replaced the misleadingly named `loadedUtilsScript` with a `Deferred` that gets resolved when the AJAX call for the utils script has completed. This way if there are additional calls to `loadUtils` after the first one, the passed `Deferred`s will only be resolved when the loading has finished (instead of immediately, as before).

I'm not sure whether `loadedUtilsScript` was supposed to be used by user code (I found no mentions in the wiki), so I removed it, but of course it could be added back for compatibility...


jackocnr:
Thank you for this. Yes I agree `loadedUtilsScript` was a misleading name, and agree `utilsLoading` is better. But I'm confused by the need for a new Deferred object - surely the one we already have is enough? No matter how many times you initialise the plugin, and on how many inputs, we should only be loading the utils script once, and that is what the existing deferred object should be tracking, no?

In the issue you say:

> A quick workaround for only this scenario is to _not_ pass the `utilsScriptDeferreds` of instances when calling `loadUtils` from `_initRequests`, because these will be resolved anyway when `handleUtils` is called

I agree with this. AFAICS this is all we need to do.

> but this is not a complete fix (manual calls to `loadUtils` would still work incorrectly).

I don't understand this bit. With manual calls to `loadUtils` you're not passing the `utilsScriptDeferred` object anyway, so it's irrelevant, no?

> I'm not sure whether `loadedUtilsScript` was supposed to be used by user code

Nah I've never said it's safe to use, so fine to change.

molnarm:
Hi,
It's been a while since I looked into this so I don't remember everything, but in any case, I added a test for the issue, so when in doubt, you can just run it and see whether it's fixed or not.

> AFAICS this (not passing deferreds from _initRequests) is all we need to do.
> With manual calls to loadUtils you're not passing the utilsScriptDeferred object anyway, so it's irrelevant, no?

If that's **really** true, than you may be right, and we don't need a second Deferred indeed, but can we count on that? The _loadUtils_ method is in the documentation - though without the second parameter, but anyone can look it up and it makes sense to pass a deferred to be notified when the loading is done, so I wouldn't be surprised if people did that.

jackocnr:
Hmm it's a good point that people who call `loadUtils` manually should be able to know when it is ready, but in this case I think we should just return the `$.ajax` call, which returns a Deferred anyway (or an object that implements the same API, so you can do `$.ajax(...).done(...)`, so if we return that object, devs will be able to call `$.fn.intlTelInput.loadUtils(...).done(...)`). What do you think?

molnarm:
I like that idea, it would make the code simpler too.

jackocnr:
Hey @molnarm, I'm ready to do a major release - any chance you could push this over the line so we can include it?

jackocnr:
Thank you so much for your input @molnarm. In the end I implemented it myself - let me know what you think: https://github.com/jackocnr/intl-tel-input/commit/295aa0d468b1d7cfcfb184b1d24fe298a37ac6cd

molnarm:
@jackocnr Sorry, I was on holiday, then trying to catch up with work. I haven't tested it now but looks OK, thanks for fixing it.

PR #722, opened 2018-04-03
Integrate css minification

joshua-s:
It would be really nice to have minified css for use in production.

@jackocnr can you review this please?


jackocnr:
Thanks!

PR #743, opened 2018-06-05
Fix for chrome: Change autocomplete to "nope"

woeterman94:
Set show autocomplete to "nope" (false and no seem to be ignored by chrome)


jackocnr:
Thanks for submitting this, but weirdly when I was testing your changes I noticed that with `autocomplete="nope"`, the input actually **did** start autocompleting, and when I changed it back to `autocomplete="off"` it stopped again. Tested in Chrome v67.0.3396.79 on MacOS. 

You can test this yourself on the demo site intl-tel-input.com and opening devtools and changing the attribute manually. I'm going to close this PR for now, and let's continue the discussion in the Issue thread.

PR #748, opened 2018-06-12
Use hiddenInput when more than one phone input in form

hamidyousefi:
@jackocnr please check these changes. I saw your last message in the issue tracker, but then I realized sometimes we don't set name for main input[type=tel] field! and just want to use the `hiddenInput` value (like demo.html inside the repo), because of that I made some changes on my own solution and add some conditions.

If you saw it is necessary to follow your last comment, please send a message and I will fix it exactly like that.


jackocnr:
Thanks for this. It looks good, but I just have two bits of feedback:

1. I would prefer to keep things simple for now, and just say in the readme that the hidden input will take the main input's name, and in the code to just assume the main input has a name.
2. I don't think we need to update the main input's name for them - I think we can leave this to them (again, we should mention this in the readme), and then that has the added benefit of giving them the flexibility to call it whatever they want (maybe they have their own naming convention and don't want to call it "main").

jackocnr:
Closing due to inactivity. I will add this myself.

PR #760, opened 2018-07-04
Fix typo in destroy method that causes popup to not close on destroy.

JvJefke:
This PR contains a simple typo in the code that causes the destroy method not to close the dropdown.


jackocnr:
Good spot, thanks! If you can remove the other (non-JavaScript) changes from your commit then I'll be happy to merge.

jackocnr:
Could we also get rid of that package-lock.json change, and squash the commits into one please?

JvJefke:
done

JvJefke:
It might be handy for you guys to write all this down in the contribution guide. Otherwise you'll have to ask this stuff everytime.

jackocnr:
Thank you so much!

PR #763, opened 2018-07-13
[Feature] Add an option for a custom user-agent string

fartek:
Added an option to use a custom user-agent string for testing whether the user device is mobile or not. Setting this option will override the default user-agent string.
This feature was added because the company I work with has a hybrid mobile app with a custom user agent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,10 @@ Allows to translate the countries by its given iso code e.g.:
 { 'de': 'Deutschland' }
 ```
 
+**customUserAgent**
+Type: `String` Default: `null`
+Use a custom user-agent string for testing whether the user device is mobile or not. This will override the default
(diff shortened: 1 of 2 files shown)


jackocnr:
Hi there, thanks for this. Can I ask what your custom user agent string is? Can't you just add one of the words we look for in the regex so it will match with the current regex?

fartek:
@jackocnr hey, I don't remember exactly right now what the string is but I know it cannot be changed because it's from a legacy project that is a couple of years old and probably has a few direct string comparisons sadly.

jackocnr:
Ok well I'm afraid I'm not going to merge this right now. I'm trying really hard to keep this plugin to be as simple and lightweight as possible, and I don't think this feature will be used by many people at all. Thanks again though.

PR #795, opened 2018-09-28
Optimisation: keep reference to activeItem and highlightedItem

### LINKED ISSUE #789: Optimisation: keep reference to activeItem and highlightedItem
If we keep a `this.activeItem` and a `this.highlightedItem` then we can simplify all of the logic around updating the active/highlighted item, which currently does a lot of filtering through every list item (~250) each time.

kbarralon:
Fix #789

Change the way we get the current active and highlighted item by keeping them in class attributes `this.activeItem` and `this.highlightedItem`.


jackocnr:
Amazing! Thank you so much.

PR #798, opened 2018-10-09
Do not show orphaned "+" when no dial code is available

nie7321:
I would like to use a default country of "none". The example showing me how to manipulate the country data got me 99% of the way there, but with one cosmetic problem:

image

Similarly, if I use `""` as the dial code, it shows a lonely "+".

This change checks the dial code when rendering that `<span>` and skips it.

My testing thus far hasn't shown this to cause any errors elsewhere, and there shouldn't be any impact on normal use-cases, as this would only come into play if you chose to add a nonsense country.

If you need anything else (jasmine test, `examples/` entry, etc), please let me know.

## Example
```js
var countryData = window.intlTelInputGlobals.getCountryData(),
  input = document.querySelector("#phone");

countryData.unshift({
  name: 'Please Select...',
  iso2: 'none',
  dialCode: null
});

window.intlTelInput(input, {
  initialCountry: "none",
  preferredCountries: ['cn', 'br', 'in', 'us'],
  separateDialCode: true,
  utilsScript: "../../build/js/utils.js"
});
```

image


jackocnr:
This is a clever hack, but it is a hack - this is not how the country data was designed to work, and I think it's likely there will be unexpected consequences.

I would ask: what is your goal here? Do you want the input to show the grey square by default? Don't you think that's confusing? Especially with no placeholder.

PR #817, opened 2018-11-06
ci: test Node.js 8, 10 and 11

DanielRuf:
Test the current LTS branches too. See https://github.com/nodejs/Release/blob/master/README.md

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
 language: node_js
 node_js:
   - "node"
+  - "10"
+  - "8"
+  - "6"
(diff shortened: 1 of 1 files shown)


DanielRuf:
Only Node.js >= 8 is supported which we should make clear in the `engines` field in package.json.

https://travis-ci.org/jackocnr/intl-tel-input/builds/451258484

jackocnr:
Thanks!

DanielRuf:
> Only Node.js >= 8 is supported which we should make clear in the `engines` field in package.json.

Should I create a separate PR for this?

jackocnr:
That would be great, thank you.

PR #820, opened 2018-11-09
chore: at least Node.js 8 is required

DanielRuf:
This will be a breaking change for users who use intl-tel-input on a platform with Node.js < 8 so we should consider to bump the major version of intl-tel-input.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,9 @@
     "region-flags": "^1.1.0",
     "time-grunt": "^1.4.0"
   },
+  "engines": {
+    "node": ">= 8"
+  },
   "repository": {
     "type": "git",
     "url": "https://github.com/jackocnr/intl-tel-input.git"


jackocnr:
Hi @DanielRuf, thank you for putting this together, and apologies in the delay in getting back to you.

Please can you clarify why Node 8 is required? Is there a feature that we're using where you can link to some documentation stating that it is only available/possible in this particular version?

Thanks again.

DanielRuf:
It is because of dependencies.

```
error grunt-contrib-jasmine@2.0.3: The engine "node" is incompatible with this m
odule. Expected version ">=8.0.0". Got "6.16.0"
error Found incompatible module
```

We would see this issue if we would test Nodejs 6 on CI which is still supported until April 2019.

See https://github.com/nodejs/Release/blob/master/README.md

As we only test 8, 10 and stable, we can only ensure that it will work with these versions.

jackocnr:
Hmm if it's just required to run the tests then I don't think it makes sense to add this as an explicit requirement - the vast majority of people who install and use this plugin will never need to run the tests.

DanielRuf:
What about us contributors?

DanielRuf:
What about the contributors and projects who use yarn and install the devDeps?

DanielRuf:
If you think it's sufficient, then let's close this PR.

PR #822, opened 2018-11-14
Fixed 765: scrolling country list on iOS uses inertia scrolling now

bartoszhernas:


jackocnr:
Thanks for this - if you could just build the CSS (`npm run build:css` should do it) then I'll merge it.

bartoszhernas:
Done

jackocnr:
Thanks again

PR #826, opened 2018-11-21
getFullNumber added code

romgom:
- added checker if user inputted number starts with +{dialCode}, remove if true
- return "Invalid"  if user inputted number length not equal to country's example number length (spaces trimmed).

### DIFF
--- a/build/js/intlTelInput.js
+++ b/build/js/intlTelInput.js
@@ -941,6 +941,7 @@
             }, {
                 key: "_updatePlaceholder",
                 value: function _updatePlaceholder() {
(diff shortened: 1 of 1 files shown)


jackocnr:
Hi there,

Thanks for this, but I'd definitely want to discuss it before merging. For both points: please can you explain why we need that change?

And if we decide to go ahead with the feature, you would need to make the changes in the src/ directory, and then run the build task (see the Contributing section of the readme) to update all of the JS files in the build/ directory. Cheers.

jackocnr:
Closing due to inactivity.

Issue #823, opened 2018-11-14
iti returns a number including country code twice


AHgPuK:
intlTelInput.getNumber returns a phone number with double country code
(Reproduced from 14.0.0 up to 14.0.6)

### Steps to reproduce
1.  Enter +78639222222
2.  Press "Submit"

### Expected behaviour
iti.getNumber should return +78639222222

### Actual behaviour
+7+78639222222

### Initialisation options
separateDialCode: true, initialCountry: 'auto', geoIpLookup included
preferredCountries: [
			'ae'
		],
		separateDialCode: true,
		initialCountry: 'auto',
		utilsScript: '/js/utils.js',
		autoPlaceholder: 'polite',
		geoIpLookup: function () {}

jackocnr:
Well the idea of `separateDialCode` is that the user doesn't need to type the dial code. But I suppose it would be worth handling the situation if they do. I would be open to a pull request to fix this - namely to check in `_getFullNumber` if the `val` starts with a plus, in which case don't add the dial code: https://github.com/jackocnr/intl-tel-input/blob/master/src/js/intlTelInput.js#L1160

jackocnr:
Fixed in v14.0.8.

PR #840, opened 2018-12-17
Fix for doubled dial code when separateDialCode is enabled

mirceaciu:
A quick fix for the doubled dial code of issue [#823](https://github.com/jackocnr/intl-tel-input/issues/823)

I've extended `_getFullNumber` to check for when a phone number starts with the dial code.

As a result this input:
image
 will output `+4071234567`


iMojtaba:
Please merge this branch ...

mirceaciu:
I've updated the PR as per recommendations.

side note: looking at the _startsWith method
```
  _startsWith(a, b) {
    return (a.substr(0, b.length).toLowerCase() === b);
  }
```
shouldn't `b` also be converted to lowercase so that apples are compared to apples?
  
 `return (a.substr(0, b.length).toLowerCase() === b.toLowerCase());`

Also [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) recommends using `substring` instead of `substr`. What's your take on this?

src/js/intlTelInput.js
```diff
@@ -1155,7 +1155,7 @@ class Iti {
 
     // normalized means ensure starts with a 1, so we can match against the full dial code
     const normalizedVal = (numericVal.charAt(0) === '1') ? numericVal : `1${numericVal}`;
-    if (this.options.separateDialCode) {
+    if (this.options.separateDialCode && val.charAt(0) !== '+') {
```
jackocnr: 👌 perfect, thank you

mirceaciu:
I'll take a better look at the code and look for `substr` and will make a PR on that(and maybe consider the _startsWith change then, if needed)

jackocnr:
Brilliant thank you. I will sort out these build conflicts and get this merged over the weekend.

jackocnr:
Merged. Thanks again.

PR #843, opened 2018-12-20
Upgrade libphonenumber to v8.10.2

dcyriller:
This would fix a bug regarding some vietnamese numbers. They were declared as invalid whereas they are valid.


dcyriller:
ping @jackocnr

jackocnr:
Hi! Thank you so much for putting this together. I'm afraid I was off cycling around India in December, but I'm back now and I will try and review this over the weekend.

jackocnr:
Thank you, it was great!

So thanks again for putting the PR together, and thanks for the tweak to the docs too. My only feedback would be with the new line in the docs - I think it would make sense to move it up a bit, to just under the clone commands, so the idea is that it's part of the initial setup process, and then from then on every time you want to rebuild libphonenumber to a new version, you just have to go through the 3 existing steps. Then I'm happy to merge. How does that sound to you?

dcyriller:
It makes sense! I've updated the PR accordingly.

jackocnr:
Thank you so much Cyrille!

jackocnr:
Released in v14.0.7

PR #845, opened 2018-12-27
Method getInstance in intlTelInputGlobals

mittaus:
Sample include in README.md

**getInstance**  
return the `iti` instance of the HTML Element
```js
var inputElement = document.getQuerySelector('#element-id');
// retrieve instance
var iti = window.intlTelInputGlobals.getInstance(inputElement);
// working with instance retrieve
var isValid = iti.isValidNumber();
console.log(isValid);
```


jackocnr:
Hi there @mittaus, thank you so much for putting this together, and apologies for the slow response.

After looking into this approach a bit, I'm concerned about the issue raised in the top answer for this related StackOverflow question: [Can I add arbitrary properties to DOM objects?](https://stackoverflow.com/questions/4258466/can-i-add-arbitrary-properties-to-dom-objects), namely:

> It's not recommended as it's messy and you may mess with existing properties.
> If you are looping code with `for..in` your code may break because you will now be enumerating through these newly attached properties.

As a plugin I think we have a special duty to avoid adding unexpected side effects like this to the DOM.

How about this approach: in the plugin initialisation code, we add a data attribute e.g. `data-intl-tel-input-id="1"` to the input element, then to grab the instance for that input all you'd have to do is `window.intlTelInputGlobals.instances[input.getAttribute('data-intl-tel-input-id')]`. What do you think?

jackocnr:
> the changes that would have to be made would be: [...]

Thanks for providing this. Actually I think we can simplify it and get rid of the jQuery checks because the jQuery plugin version has it's own way of handling this, with the standard `$("#phone").intlTelInput(...)`.

Also in outro.js, couldn't we just use `iti.id` as the id?

Anyway, would you be willing to update your PR to reflect these changes? I'd love to add you to the list of contributors to this project!

jackocnr:
I've just realised we will also want to remove the new data attribute in the destroy method.

If you're interested in making these changes yourself then please let me know, else I will probably have a look at it myself on Sunday. Cheers.

mittaus:
Corrected:

1. Checking of Jquery Object has been removed
2. The id property of the instance is used for the attribute in data-intl-tel-input-id
3. The attribute data-intl-tel-input-id is removed in the destroy method

jackocnr:
Thank you very much for this. I made a couple of minor tweaks, and squashed the commits into one, which is now in master: https://github.com/jackocnr/intl-tel-input/commit/42635c2ce0dcda72281a16c025f438ee35fa1158 and will be included in the next release!

Thanks again.

PR #848, opened 2019-01-04
Fixed 847: autoHideDialCode - Browser form validation message gets hidden after dial code re-added

bencergazda:


jackocnr:
Thanks but as I mentioned in the issue thread, I'd like to discuss this first. I'd be really reluctant to add this much code/complexity just to support this old feature which I'm starting to think is a bad idea anyway.

Also, according to MDN, [setCustomValidity is not supported by Safari](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/setCustomValidity).

bencergazda:
@jackocnr Ok, I understand your point of view.

> Also, according to MDN, [setCustomValidity is not supported by Safari](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/setCustomValidity).

Thanks, I haven't noticed it. I was pretty sure Safari does still lack the whole validation API. :-)

Issue #668, opened 2017-12-13
Chrome 63 is ignoring autocomplete="off"


ibarral:
Chrome 63 is ignoring autocomplete="off", and tries to suggest previous user entries.

The only solution is to assign a random string to the attribute, for example:

`autocomplete="nope"`

Source: https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

I think that we'll need a fix to that :(

jackocnr:
Thanks for raising this. As I mentioned, I would be open to a PR for this tweak! Thanks.

woeterman94:
Any update on this?
Can we still use autocomplete=off on input fields?

woeterman94:
Solution: use autocomplete="nope"

jackocnr:
@woeterman94 yup, that was in the original post haha. I'm open to a pull request :)

woeterman94:
Created one :)

jackocnr:
Strangely, on Chrome v67 (67.0.3396.79) on MacOS, having the autocomplete attribute set to "off" is working for me now (disabling autocompletion), and setting it to "nope" is not working (autocomplete is working again). Are you guys seeing the same thing?

woeterman94:
Works for me with the latest chrome build on windows.

jackocnr:
@ibarral what's your experience? (and are you Windows/Mac etc?)

tinisasa:
I'm still having this issue on inputs, forms, using autocomplete"off/nope/whatever" and/or autofill"off/nope/whatever" 
Chrome Version 66.0.3359.181 (Official Build) (64-bit)
MacOD 10.13.3

Shadosky:
It look like you can use `autocomplete="new-password"` now. 

cf : https://bugs.chromium.org/p/chromium/issues/detail?id=370363#c7

Tested on Version 67.0.3396.99 (Build officiel) (64 bits), & working

jackocnr:
@Shadosky unfortunately that suggestion was from 4 years ago! I'm closing this for now, as it seems it is impossible to fix!

juan0087:
I'm on Chrome Version 68.0.3440.106 (Official Build) (64-bit) and I was able to hide it using       autoComplete="nope" but only on mac osx. On the same browser in a Windows 10 machine it still shows the autocomplete.

justjuank:
[...]

PR #849, opened 2019-01-07
Change autocomplete attribute from off to nope

kanenas:
Although autocomplete **IS** set to **OFF** browsers (Google Chrome v.71.0.3578.98 sure does this) seem to ignore the value, so based on https://stackoverflow.com/a/34998771/388382 by setting the value to "**nope**" will do the job.
Tested with **Google Chrome v.71.0.3578.98** & **Firefox Quantum v.64.0**


jackocnr:
Hi there, thanks for this. Did you see this conversation about this attribute? https://github.com/jackocnr/intl-tel-input/issues/668

I only want to proceed with this if there's an agreed standard that we can reference, as previously this has been a moving target, with fickle/changing support on each browser on each platform.

kanenas:
Sorry :( i didn't see the conversation #668

PR #878, opened 2019-02-27
Update README.md

landsman:
added debug info about -99 code


jackocnr:
Hi @landsman, thanks for putting this together. From a quick glance at the source code, it looks like this -99 error code only applies to two methods: `getValidationError` and `getNumberType`, so rather than add a new section to the already long readme, I'd rather just mention it in the description for those two methods. Would you be up for making that change?

jackocnr:
Closing due to inactivity.

PR #892, opened 2019-04-22
Adding focus code for dropdown list item so that element should have …

padu:
…focus state with highlight class, this would help screen reader device to identified navigated list item


padu:
@jackocnr Please review

jackocnr:
Hi there, thanks for submitting this. I'm afraid I currently have a broken wrist, so can't test this right now. Maybe next week. In the meantime, can you tell me (A) does this change the experience in any way on a regular desktop computer? And (B) instead of adjusting the tabIndex every time, perhaps when we generate the markup we could just assign a tabIndex of -1 to every list item, which apparently makes it focusable from JS code, but not from the user's tab key. What do you think?

padu:
(A) - This change the experience for screen reader users, for example  a blind person want to access our field then he can use NVDA opensource screen reader for desktop and should able to access phone field controls using tab, up/down arrow, space/enter keyboard keys and screen reader dictates screen text exposed through aria attributes . I can see we have used aria role="combobox" and related attributes very well.

On the selection side we have manged highlighting drop-down option with up/down arrow keys but missed to focus highlighted field. This is the reason blind user uses up/down key to navigate drop-down item but screen readers not able to detect highlighted field so not able to announce it.

(B) When we generate markup we can assign tabIndex = -1, but still we need to change its value on navigating to option field with up/down key. tabIndex = -1 means field is not focusable, we are explicitly making highlighted field focusable with tabIndex = 0, and calling element.focus on it to active focus to convey screen reader.

@jackocnr, Take care of your wrist, meanwhile we can have quick call for test demo on skype (padupank4).

jackocnr:
Thanks so much for the detailed response @padu, and apologies if I wasn't clear enough before - let me re-phrase. (A) I know this change is intended to benefit people using screen readers, but I was just wondering if it also had any affect on the standard desktop web experience? And (B) if you give elements a tabIndex of -1 then actually we can focus them in the JavaScript - see https://codepen.io/jackocnr/pen/MRROvM - what do you think?

padu:
A) For standard desktop web experience we have done keyboard management (up/down keys) with highlight class added. In addition we want to add actual semantic meaning along with highlight class, we can focus to element. This affects to add focus style along with highlight class as desktop experience user can easily overwrite it with CSS sudo class ":focus" . No effects other than this.

B) element with tabIndex=-1, can focused with javascript but not with keyboard tab. It just take out that element from default tab order. tabIndex = 0, make it focus-able on tab. In our case we want to dynamically control element tab order with focus.

check with https://jsfiddle.net/Pank/rgdvbtLw/, just press tab focus will move to element having tabIndex=0, skipping all the elements having tabIndex=-1

jackocnr:
Thank you. Re: (B), I don't understand why we need to support the tab key - we are calling focus() in the JavaScript, so there is no need to support the tab key as well. Also when the dropdown is open, we have a keydown listener, and when tab is pressed we close the dropdown, so it wouldn't work anyway.

padu:
I think you are right tabIndex manipulation is redundant, we can assign it tabIndex=-1 while creating dropdown items. Let me check it, I will updated code.

padu:
Done. It looks good for screen-readers. checked with NVDA on chrome/firefox.

padu:
@jackocnr, Please review code changes., I want to know if anything need to update for build failure.

jackocnr:
Apologies for the delay in reviewing this. And thanks for your work so far - it is appreciated.

I've spotted a couple issues while testing on Chrome desktop that we would need to address before merging this.

Firstly upon opening the dropdown, there is a weird blue line between the first two items:

Screenshot 2019-05-12 at 14 09 20

This is due to the top item having focus, and so Chrome is adding its standard blue outline, but for some reason the top and sides are cutoff. In this case I would probably recommend adding `outline: none;` to the `.country` CSS styles.

Secondly, I'm seeing a weird scroll effect when I move the mouse up and down around the bottom of the dropdown list - somehow it has started jumping around. It must be to do with the new focused state - any ideas on how to prevent this?

padu:
@jackocnr Thanks for review.
Added outline none to style css.

Ideally scroll updates to focused element and in our case if at top or bottom if any element is partially hidden scroll updates to take hovered country item in to view.

I think this should be expected behavior, otherwise user see partially hidden but highlighted element.
I agree it looks weird but if we carefully observed, scroll is just taking highlighted list item in to view so that user can see it with ease. 

Also as we have focused state applied 

If we don't want this to happen. On mouse-over we can manage highlighting with CSS except event callback '_highlightListItem' to call on mouse in.

jackocnr:
> if we carefully observed, scroll is just taking highlighted list item in to view so that user can see it with ease

I see your point, but unfortunately it feels buggy to me. And the more I think about it, the more I think it doesn't actually make sense to highlight a hovered country in the same way as we do when you highlight a country with the arrow keys. When you highlight a country with the arrow keys, that's when it makes sense to be able to select it by hitting enter. But if you highlight a country with the mouse, to me that is only for if you then click to select it - enter shouldn't work in this case.

But yeah for now, maybe there is a way to prevent this focus from happening on hover?

padu:
We can add conditional focus taking one more parameter in _highlighListItem, Take a look.
Also I have commented scrollTo function as it is redundant with auto scroll on element focus.

jackocnr:
Looks good! Nice work.

Re: scrollTo, I think we don't need it when hitting up/down, as I think focus handles that perfectly, so let's remove it in `_handleUpDownKey`. I think that when you type a country name, I actually think it's a much better UX for the country to appear in the middle of the dropdown instead of at the top/bottom, so in `_searchForCountry` I want to keep the call to scrollTo. And after thinking about this, I actually think that when you first open the dropdown, it makes the most sense for the highlighted country to be in the middle then too, so in `_showDropdown` can we actually keep the call to scrollTo, and add `true` for the second argument there, so that it scrolls the country to the middle.

Do you agree?

padu:
Got it, I Agree.
commited changes as you said we want the selected country to be the middle for search and to be shown first time. removed scrollTo in up down arrow keys logic.

jackocnr:
This is looking really great, thank you.

Lastly, just some admin:

1. formatting: please remove brackets from single line if statements and add space to `if(shouldFocus)`
2. build the js with `grunt js`
3. squash all of your commits into one

Then I will merge! Thanks again.

padu:
Committed changes for formatting and build javascript.

All my commits were in forked master branch, So don't know how I can squash all in to one.
I did "git rebase -i HEAD~[10]" for all my previous commits.

4lex96:
Hey guys, thanks for the incredible work you've done. @jackocnr how's your wrist?

Please could you accept the merge requested by @padu ? Because at this moment the intl-tel-input of the demo page is not accessible.

Best regards,
Alex

jackocnr:
I manually squashed the commits and merged. But your name is still on the commit @padu, thank you so much. Released in v15.1.0.

padu:
Thanks @jackocnr

Issue #885, opened 2019-04-04
CSS display Bug since v14.0.10


m-dl:
### Steps to reproduce
I just updated from v14.0.7 to v14.0.10+, and I noticed a bug in display (padding, and width).
It seems to be since this commit, in v14.0.10 : [here](https://github.com/jackocnr/intl-tel-input/commit/48c95a850ac71c16e9ded5d3839eb25498275075)

### Expected behaviour
This is what should be displayed as on previous versions : 
[https://imgur.com/WCRv5N6](https://imgur.com/WCRv5N6)

### Actual behaviour
This is what is displayed, arrow is not well positionned, and when default number is set, there is no padding on the input :
[https://imgur.com/GXhw5Q4](https://imgur.com/GXhw5Q4)
[https://imgur.com/YezBrIv](https://imgur.com/YezBrIv)

### Initialisation options
Tested on Firefox 66, and Chrome 73.

Some screens of the CSS console : 
[https://imgur.com/749dmaV](https://imgur.com/749dmaV)
[https://imgur.com/dTB9ZPs](https://imgur.com/dTB9ZPs)
[https://imgur.com/5oPaCpF](https://imgur.com/5oPaCpF)

```
this.intlPhone = intlTelInput($("#" + this.id)[0], {
                utilsScript: require("intl-tel-input/build/js/utils.js"),
                initialCountry: "auto",
                separateDialCode: true,
                nationalMode: true,
                formatOnDisplay: true,
                autoPlaceholder: "polite",
                preferredCountries: [],
                geoIpLookup(success, failure) {
                    $.get("https://ipinfo.io", function() {}, "jsonp").always(
                        function(resp) {
                            let countryCode =
                                resp && resp.country ? resp.country : "";
                            success(countryCode);
                        }
                    );
                }
            });
```

crabnky:
+1

malecki-se:
+1

malecki-se:
@jackocnr The problem is probably with wrong build in npm, because this problem with spaces doesn't exist in this repo

jackocnr:
Works for me: https://codepen.io/jackocnr/pen/mgBYXG (tested against the latest version)

I'm guessing this is a clash with your own custom CSS. If you can reproduce the issue in a codepen with the minimal code required then I'll be happy to take a look.

malecki-se:
@jackocnr You can check it on: https://codesandbox.io/s/vjppprpzl0?from-embed

[...]

PR #906, opened 2019-06-05
Issue #885

mac89:
Fixes one of the issues reported in #885. When the input was not visible on initialization the `offsetWidth` could not be calculated.


jackocnr:
Hi @mac89, thanks for putting this together, but I need to be able to reproduce the issue first before I would consider what the best fix would be - see my comment in the issue.

Issue #912, opened 2019-06-07
Showing Russian flag for Kazakhstan number


jyotsnagaddam:
### Steps to reproduce
1. Please take any Kazakhstan number online
2.  Go to the demo page https://intl-tel-input.com/node_modules/intl-tel-input/examples/gen/is-valid-number.html
3.  And enter the number

### Expected behaviour
Should show Kazakhstan country flag

### Actual behaviour
But showing Russian Flag

### Initialisation options
List any options you're using e.g. utilsScript or preferredCountries

jyotsnagaddam:
Note: Both Russia and Kazakhstan share same international dial code (+7)

Based on the below specification and data:
 // So each country array has the following information:
    // [
    //    Country name,
    //    iso2 code,
    //    International dial code,
    //    Order (if >1 country with same dial code),
    //    Area codes (if >1 country with same dial code)
    // ]


[
    "Kazakhstan (Казахстан)",
    "kz",
    "7",
    1
  ],

**As area codes are missing above, it seems it showing the Russian flag based on the priority set**

jackocnr:
Well spotted. So currently we only support "area codes" for countries that share the international dial code of +1, but actually we also need to support them for +7 (shared by Russia and Kazakhstan), as well as the following others: +39, +44, +47, +61, +212, +262, +358, +590, +599, which are all used by more than one country/territory.

The first step is to collect the list of area codes for each of these countries - is that something you could do, and just post them here for reference?

jyotsnagaddam:
I have attached the "area codes" for the countries that share the international dial code.

Note:
- Some countries share common area codes and cannot be separated. (Ex: Guadeloupe, Saint Barthelemy, Saint Martin)
- Some Islands are uninhabited or with temporary inhabitants.
- Vatican City is with in the Rome and shares same area code (06) but the numbers usually starts with 698.

References: 
https://github.com/google/libphonenumber/blob/master/resources/PhoneNumberMetadata.xml

Cross References:
- https://www.itu.int/oth/T0202.aspx?lang=en&parent=T0202
- https://en.wikipedia.org/wiki/List_of_country_calling_codes
- https://www.howtocallabroad.com/

Better have it someone review the attachment.
[Area Codes.xlsx](https://github.com/jackocnr/intl-tel-input/files/3270693/Area.Codes.xlsx)

jackocnr:
This is great thanks!

[...]

PR #914, opened 2019-06-10
Update data.js

jyotsnagaddam:
Showing Russian flag for Kazakhstan number (#912)
Added area codes for countries that share same international dial codes.
1. Western Sahara, +212
2. Mayotte, +262
3. Aland Islands, +358
4. Vatican City, +39
5. Guernsey, +44
6. Isle of Man, +44
7. Jersey, +44
8. Svalbard and Jan Mayen, +47
9. Caribbean Netherlands, +599
10. Kazakhstan, +7

### DIFF
--- a/src/js/data.js
+++ b/src/js/data.js
@@ -212,7 +212,8 @@ var allCountries = [
     "Caribbean Netherlands",
     "bq",
     "599",
-    1
+    1,
+    ["3", "4", "7"]
   ],
   [
     "Cayman Islands",
@@ -470,7 +471,8 @@ var allCountries = [
     "Guernsey",
     "gg",
     "44",
-    1
+    1,
+    ["1481", "7781", "7839", "7911"]
   ],
   [
     "Guinea (Guinée)",
@@ -541,7 +543,8 @@ var allCountries = [
     "Isle of Man",
     "im",
     "44",
-    2
+    2,
+    ["1624", "74576", "7524", "7924", "7624"]
   ],
   [
     "Israel (‫ישראל‬‎)",
@@ -570,7 +573,8 @@ var allCountries = [
     "Jersey",
     "je",
     "44",
-    3
+    3,
+    ["1534", "7509", "7700", "7797", "7829", "7937"]
   ],
   [
     "Jordan (‫الأردن‬‎)",
@@ -581,7 +585,8 @@ var allCountries = [
     "Kazakhstan (Казахстан)",
     "kz",
     "7",
-    1
+    1,
+    ["33", "7"]
   ],
   [
     "Kenya",
@@ -717,7 +722,8 @@ var allCountries = [
     "Mayotte",
     "yt",
     "262",
-    1
+    1,
+    ["269", "639"]
   ],
   [
     "Mexico (México)",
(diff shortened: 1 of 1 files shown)


jackocnr:
This looks perfect, thank you very much.

Please could you try [building the JS](https://github.com/jackocnr/intl-tel-input/blob/master/.github/CONTRIBUTING.md#changes-to-the-plugin) and check that everything is still working, and that the tests still pass?

jackocnr:
I've just checked and the tests pass, and everything seems to be working, but just one issue: we need to update this line as the max dial code length has now grown from 4 to 7: https://github.com/jackocnr/intl-tel-input/blob/master/src/js/intlTelInput.js#L1106

jyotsnagaddam:
Yes auto-generation is a right way to address the issue. I did the changes as proposed. I just tested on my local copy and changes are looking fine. Please validate.

jyotsnagaddam:
I have a question regarding 'separatedialcode' option. It seems only for NANP countries who have single area code will have the dialcode as +1 followed by area code. For other countries like US and CA dialcode will show as '+1'

But for all other countries that share same international dialing code, we are not showing the area code followed by country code even if the country has single area code when we set the 'seperatedialcode'.

Is my understanding correct?

jackocnr:
> I did the changes as proposed. I just tested on my local copy and changes are looking fine. Please validate.

Did you push the changes? I'm not seeing them in this PR.

> I have a question regarding 'separatedialcode' option. It seems only for NANP countries who have single area code will have the dialcode as +1 followed by area code. For other countries like US and CA dialcode will show as '+1'

> But for all other countries that share same international dialing code, we are not showing the area code followed by country code even if the country has single area code when we set the 'seperatedialcode'.

> Is my understanding correct?

Yes that is correct. Currently for some NANP countries we include the area code in the dial code field in data.js, e.g. [+1242 for Bahamas](https://github.com/jackocnr/intl-tel-input/blob/master/src/js/data.js#L92). Now that I think about this, I think it would be better to keep the area codes separate even in cases where there is only one, because even in those cases, the official international dial code is still just +1. But I think don't worry about that in this PR - I think we can refactor those ones in another PR. Cheers.

jyotsnagaddam:
My bad, I thought I created the pull request yesterday itself. It seems, I just saw the preview changes and forgot submitting the pull request. I just did it, can you please check?

jyotsnagaddam:
> > I did the changes as proposed. I just tested on my local copy and changes are looking fine. Please validate.
> 
> Did you push the changes? I'm not seeing them in this PR.
> 
> > I have a question regarding 'separatedialcode' option. It seems only for NANP countries who have single area code will have the dialcode as +1 followed by area code. For other countries like US and CA dialcode will show as '+1'
> 
> > But for all other countries that share same international dialing code, we are not showing the area code followed by country code even if the country has single area code when we set the 'seperatedialcode'.
> 
> > Is my understanding correct?
> 
> Yes that is correct. Currently for some NANP countries we include the area code in the dial code field in data.js, e.g. [+1242 for Bahamas](https://github.com/jackocnr/intl-tel-input/blob/master/src/js/data.js#L92). Now that I think about this, I think it would be better to keep the area codes separate even in cases where there is only one, because even in those cases, the official international dial code is still just +1. But I think don't worry about that in this PR - I think we can refactor those ones in another PR. Cheers.

Got it. Thanks.

jackocnr:
Thanks so much for your help here 👍

PR #916, opened 2019-06-15
Update intlTelInput.js

jyotsnagaddam:
auto-generating max dial code length, to protect us if we end up adding even longer ones in the future. Also technically the number could vary between instances, depending on the onlyCountries setting, so it would be great to generate one per instance, during the plugin init code, and store it as an instance variable. 
-- As proposed by the author --


jackocnr:
Thanks!

PR #925, opened 2019-06-24
Add support for SSR by checking if window is available.

nicolasigot:
Hello,
As I am using this library I noticed that window usage is not protected. This is an issue when the lib in included in bundle which is used server side for rendering.
The usage is to be kept only for client side rendering.
Thanks,
Nicolas

### DIFF
--- a/src/js/intro.js
+++ b/src/js/intro.js
@@ -1,9 +1,11 @@
-// wrap in UMD
-(function(factory) {
-  var intlTelInput = factory(window, document);
-  if (typeof module === "object" && module.exports) module.exports = intlTelInput;
-  else window.intlTelInput = intlTelInput;
-}(function(window, document, undefined) {
-  "use strict";
+// We check if we are in browser
+if (typeof window !== 'undefined') {
+  // wrap in UMD
+  (function(factory) {
+    var intlTelInput = factory(window, document);
+    if (typeof module === "object" && module.exports) module.exports = intlTelInput;
+    else window.intlTelInput = intlTelInput;
+  }(function(window, document, undefined) {
+    "use strict";
 
-  return (function() {
+    return (function() {
--- a/src/js/outro.js
+++ b/src/js/outro.js
@@ -1,10 +1,11 @@
-    // convenience wrapper
-    return function(input, options) {
-      var iti = new Iti(input, options);
-      iti._init();
-      input.setAttribute('data-intl-tel-input-id', iti.id);
-      window.intlTelInputGlobals.instances[iti.id] = iti;
-      return iti;
-    };
-  })();
-}));
+      // convenience wrapper
+      return function(input, options) {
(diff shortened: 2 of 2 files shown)


jackocnr:
Thanks for this. I have to say I'm not super familiar with server side rendering. Could you link me to an example of another popular JavaScript lib which takes this approach? Thanks.

nicolasigot:
Hi @jackocnr ,

Thanks for your feedback. To give you a bit of context, I am using your plugin within my Vue application. Vue allows the application to use SSR for rendering on server side. Here is a link if you are interested to understand it a bit more: https://ssr.vuejs.org/#what-is-server-side-rendering-ssr as well as the notion of universal code (independent of the platform it is executed). Basically document and window only exist in browser and not server side.

Within my Vue application, I import the library as part of the bundle (built by webpack) within my component but I only use it within specific method (lifecycle of the component) which is only executed in the browser (where window and document are executed). The issue here is that importing the library execute the code and then tries to access window and document by default. This is failing on SSR side as window is not defined.

Basically to have your plugin compatible with SSR, you should make sure that when the script is executed, it does not access window or document. This can still be done but within methods which are called in the browser context.
The fix I did is the easiest way meaning that I protect the code from being executed during server side rendering.
It is fine as I do not expect the manipulation of DOM you do with your plugin to be rendered server side (it has no added value for SEO for example). Your plugin is pure UI / UX enhancement. If we would like to make your plugin compatible for SSR for Vuejs for example we should create a specific directive which manipulate the virtual DOM (available on server side and client side). But that's not the goal of your plugin.

I hope this clarifies a little bit what you are expecting. Let me know if you would need more information or updates from me.
Thanks!

jackocnr:
Thanks so much for writing up that clarification.

I would however still like to see an example of a popular lib adopting the same approach, to be sure this is the appropriate solution to this problem, and to confirm that it is useful to a wider audience. If we were unable to find an example of this, it would make me wonder if there might be another solution to your particular problem, e.g. a build step like a webpack configuration setting or something like that. Hope you understand.

nicolasigot:
Hello @jackocnr ,
Just to let you know that I found an other workaround to support your library on SSR side. Basically I do not load the code on SSR side.
This is how I do it:
```
{
  test: /\.js$/,
  include: [ // use `include` vs `exclude` to white-list vs black-list
              require.resolve("intl-tel-input") // intl-tel-input is accessing window by default so must be included in bundle and associated to null-loader
          ],
  loader: 'null-loader'
}
```
I still think that you should not access window global variable on load of the library. It can be accessed when we are creating the object on an element of the DOM (which is to be used on client side methods only in the application).
Thanks for your feedback and library!
Regards,
Nicolas

jackocnr:
Ok thanks for posting that here - perhaps it will be helpful to others. I'll close this now then.

PR #962, opened 2019-09-04
Fixed the issue of `setNumber` not working in Microsoft Edge if the input field doesn't set `maxlength` value

manawat:
Calling `setNumber` in Microsoft Edge on Windows will always set the input field to *blank* if that input field doesn't set `maxlength` attribute.

This is because, somehow, the command `this.telInput.getAttribute('maxlength');` inside `_cap` method returns _-1_ in Edge instead of _null_ like the other browsers.

To reproduce this problem, just run this fiddle in Microsoft Edge ( on Windows ) comparing to running it on other browser like Chrome or Firefox.
https://jsfiddle.net/narok119/g9xve23d/ 

This problem doesn't occur in Microsoft Edge on Mac tho, just on Windows.


jackocnr:
Hmm that's strange, I can't reproduce that. If I open Edge v18 (on Windows 10) on Browserstack, and open the demo site http://intl-tel-input.com and open a console and type `document.querySelector('#phone').getAttribute('maxlength')` then I get `null`.

I also can't find any reference to this `-1` quirk online - can you point me to anything?

manawat:
Sorry. I think I've put the wrong version of jsfiddle in the comment.

To reproduce this, we still have to put the `maxlength` attribute but without setting value to it.

I've also updated the jsfiddle, you can now retry this fiddle again on BrowserStack's Edge comparing with other browsers.
https://jsfiddle.net/narok119/g9xve23d

jackocnr:
But why would you have a maxlength attribute without giving it a value? That doesn't make sense.

manawat:
> But why would you have a maxlength attribute without giving it a value? That doesn't make sense.

I know right.

We actually got this problem while using a drag-and-drop web framework from Temenos, it's being used by many banks.

This framework generate empty `maxLength` attribute to the all the text field if we don't explicitly specify it.

So when we try to apply this library to the text fields generated from this framework, we got all of them blank in Edge.

jackocnr:
Hmm that sounds like an issue with another 3rd party library. I don't want to add weight/complexity to this plugin just to support workarounds for issues with other libraries, sorry.

PR #978, opened 2019-10-09
Update libphonenumber to v8.10.20

omarestrella:
Recently, the Philippines updated their number formats. Google's libphonenumber was recently updated to accommodate this change. This brings the library up to date so that number validation can happen correctly.

To check against current version of libphonenumber in intl-tel-input, you can use a new valid number like: `+63288643300`. It will fail validation.


jackocnr:
Thanks so much!

jackocnr:
Released in v16.0.5.

PR #988, opened 2019-10-18
Add navigation with arrows support for IE

reiherric:


jackocnr:
Thanks so much Reinaldo!

reiherric:
oh, my pleasure. I did it because we are using the plugin and
we needed to  keep it accessible for blind users in all browsers.
You do great work!
Congrats!
________________________________

Issue #937, opened 2019-07-16
getValidationError returns 5


princefishthrower:
### Steps to reproduce
1.  I call ini.isValid() first
2.  If false, I check the validation error to get more info
3.  validation error is 5

Example number is +4112312312123

### Expected behaviour
getValidationError() should return a number in the range 0-4 as described in the docs and source code:

```
// copied this from i18n.phonenumbers.PhoneNumberUtil.ValidationResult in https://github.com/googlei18n/libphonenumber/blob/master/javascript/i18n/phonenumbers/phonenumberutil.js and again put the keys in quotes.
// Also: added NOT_A_NUMBER to match i18n.phonenumbers.Error.NOT_A_NUMBER
var validationError = {
  "IS_POSSIBLE": 0,
  "INVALID_COUNTRY_CODE": 1,
  "TOO_SHORT": 2,
  "TOO_LONG": 3,
  "NOT_A_NUMBER": 4
};
```

Note: output from libphonenumber test site is:

```
****Parsing Result:****
{"country_code":41,"national_number":12312312123,"raw_input":"+4112312312123","country_code_source":1}

****Validation Results:****
Result from isPossibleNumber(): false
Result from isPossibleNumberWithReason(): 
Note: numbers that are not possible have type UNKNOWN, an unknown region, and are considered invalid.

****ShortNumberInfo Results:****
Result from isPossibleShortNumber: false
Result from isValidShortNumber: false
Result from isPossibleShortNumberForRegion: false
Result from isValidShortNumberForRegion: false

****Formatting Results:**** 
E164 format: invalid
Original format: +4112312312123
National format: 12312312123
International format: invalid
Out-of-country format from US: invalid
Out-of-country format from Switzerland: invalid

****AsYouTypeFormatter Results****
Char entered: + Output: +
Char entered: 4 Output: +4
Char entered: 1 Output: +41 
Char entered: 1 Output: +41 1
Char entered: 2 Output: +41 12
Char entered: 3 Output: +41123
Char entered: 1 Output: +411231
Char entered: 2 Output: +4112312
Char entered: 3 Output: +41123123
Char entered: 1 Output: +411231231
Char entered: 2 Output: +4112312312
Char entered: 1 Output: +41123123121
Char entered: 2 Output: +411231231212
Char entered: 3 Output: +4112312312123
```

### Actual behaviour
getValidationError returns 5

### Initialisation options
config options as follows:
```
window.intlTelInput(input, {
[...]

PR #989, opened 2019-10-20
Updated validationError codes and function getValidationError to match libphonenumber's API

princefishthrower:
This should solve https://github.com/jackocnr/intl-tel-input/issues/937

I added the updated libphonenumber.ValidationResults IS_POSSIBLE_LOCAL_ONLY and INVALID_LENGTH, subsequently bumped the custom NOT_A_NUMBER up to the enum value of 6

However, I couldn't seem to properly run the grunt build to get the corresponding `build/js/utils.js` to be generated. All the commands (test, build, etc) seemed to run fine in my terminal, but utils.js wasn't updated. Any ideas?

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -72,13 +72,19 @@ function getValidationError(number, countryCode) {
       return i18n.phonenumbers.PhoneNumberUtil.ValidationResult.INVALID_COUNTRY_CODE;
     }
     if (e.message == i18n.phonenumbers.Error.NOT_A_NUMBER) {
-      return 4;
+      return 6;
     }
     if (e.message == i18n.phonenumbers.Error.TOO_SHORT_AFTER_IDD || e == i18n.phonenumbers.Error.TOO_SHORT_NSN) {
       return i18n.phonenumbers.PhoneNumberUtil.ValidationResult.TOO_SHORT;
     }
     if (e.message == i18n.phonenumbers.Error.TOO_LONG) {
       return i18n.phonenumbers.PhoneNumberUtil.ValidationResult.TOO_LONG;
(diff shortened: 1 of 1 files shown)


jackocnr:
Amazing, thanks for this!

I'm afraid updating utils.js is a bit more involvedm, but should only take 5 mins, instructions here: [Updating to a new version of libphonenumber](https://github.com/jackocnr/intl-tel-input/blob/master/.github/CONTRIBUTING.md#updating-to-a-new-version-of-libphonenumber).

jackocnr:
Thanks again for this, but actually I've decided to go with a different approach.

So actually `i18n.phonenumbers.Error.IS_POSSIBLE_LOCAL_ONLY` doesn't exist - you can see the `i18n.phonenumbers.Error` object in the libphonenumber docs [here](https://github.com/google/libphonenumber/blob/master/javascript/i18n/phonenumbers/phonenumberutil.js#L69), which is different to the `i18n.phonenumbers.PhoneNumberUtil.ValidationResult` object, which you can see [here](https://github.com/google/libphonenumber/blob/master/javascript/i18n/phonenumbers/phonenumberutil.js#L965).

I think when I originally wrote this code I was trying to be clever, and make it as intuitive as possible for devs to use, but actually we've just ended up with something confusing and now broken as libphonenumber changes their enums. So now I think the best thing to do is to keep it simple and in `getValidationError` just return `e.message` directly, and add to the docs that this will be a `i18n.phonenumbers.Error` enum, and link to the libphonenumber source and then this plugin will be a lot more stable and flexible and devs can handle errors however they want.

So I'll close this PR. Thanks again.

Issue #977, opened 2019-10-07
Country code dropdown is not accessible on visually impaired users - accessibility


DimitrK:
### Steps to reproduce
1. Use the intl-tel-input library on your page  
2.  Use a screen reader (Cmd+F5 on mac)
3.  Focus on the country code input element and press down arrow.

### Expected behaviour
The assistant should announce a combobox has opened and the selected option. As you navigate and focus moves to other option, the option should get announced

### Actual behaviour
Nothing gets announced

The problem seems to be that `aria-expanded` and `aria-activedescendant` properties is passed on listbox while should be on combobox

jackocnr:
Thanks for reporting this, I would be open to a pull request that fixes it.

incohe:
i also encounter the same issue on google lighthouse audit accessibility section. upon checking the document the google provided it seems the issue is the improper use of role tag and aria-owns tag. 

By removing those tags in intlTelInput.js the error in google lighthouse audit is gone and no affect on functionality or layout 
![screenshot1](https://user-images.githubusercontent.com/55270275/68470805-c2b0d300-0257-11ea-9eef-4c3a29b5c587.PNG)
![screenshot2](https://user-images.githubusercontent.com/55270275/68470807-c3496980-0257-11ea-84ec-d7f3923f3c16.PNG)

if role tag and aria-owns is important i would like to suggest that use the proper html for the certain role like <select> and <option> tag

eunhuiii:
I have ame problem.  Have you solved this problem?

PR #1002, opened 2019-11-15
Fixes 977: Move aria-expanded and aria-activedescendant properties to combobox

RomanBaiocco:
Fixes https://github.com/jackocnr/intl-tel-input/issues/977


jackocnr:
Apologies for the delay! Thank you so much for this :)

Issue #1024, opened 2020-01-19
Allow country data to be exported separately


Kasheftin:
There's one nice file, src/js/data.js.
A lot of libraries use it. 
But since it's just a plain variable, it's not possible to import this file directly. 
Instead of using intl-tel-input they just copy the file content and append it with a single `export default allCountries`. 

Take a look: 
https://github.com/EducationLink/vue-tel-input/blob/master/src/assets/all-countries.js,
https://github.com/mihajlon97/vue-tel-input/blob/master/src/assets/all-countries.js, https://github.com/patw0929/react-intl-tel-input/blob/master/src/components/AllCountries.js  

I also have my own implementation of telephone input based on vuetify, libphonenumber-js, and I also copied data.js and flags from intl-tel-input.  

What if we add a simple grunt command for exporting and minifying data.js only?

PR #1025, opened 2020-01-19
allCountries copied to the separate data.js and data.min.js build

Kasheftin:
Build appended with data.js and data.min.js files that contain allCountries array only.


jackocnr:
Hi there. Can you tell me why you have done this?

Kasheftin:
Hi. I use data.js in my own implementation of telephone-input based on vuetify and libphonenumber-js. The last does not have country names. I need to show country flag selector, and the data from intl-tel-input works just fine for that. Previously I just copied the content of that file, but it seems to be a more true-way to load it with npm. Here I wrote about it, https://github.com/jackocnr/intl-tel-input/issues/1024.

jackocnr:
Hmm fair enough. I don't see any reason to not do this 🤷‍♂ 

But I don't think we need to use the grunt replace task in this case - I don't think anything will get replaced. So if you could remove that stuff, then I think I'm happy to merge.

Thanks.

Kasheftin:
That is strange. I can not find the reason why the tests were failed on node8. utils.js triggered an error. My update was so minor I don't believe it breaks anything.

Kasheftin:
Is it possible to rerun these tests?

Issue #1031, opened 2020-02-12
Country search does not work with localized country names


koktut:
When using localizedCountries, the country search does not work, because The _bindDropdownListeners () function filters input with the expression / ^ [a-zA-ZÀ-ÿ] $ /. test (e.key) which does not contain Russian letters

### Actual behaviour
Country search does not work

### Initialisation options
 localizedCountries: {
	'ru': 'Россия',
	'kz': 'Казахстан'
}

jackocnr:
Good catch! I'm afraid I don't have time to work on this myself right now, but would welcome a pull request.

koktut:
Done https://github.com/jackocnr/intl-tel-input/pull/1032

jackocnr:
Fixed in v16.0.11.

PR #1032, opened 2020-02-13
Fix country search (issue: #1031)

koktut:


jackocnr:
Thanks for putting this together.

I'd really rather avoid storing every country name character in memory like that. Can't we just tweak the regex to allow for Russian chars?

koktut:
I have done so to avoid similar problems with other languages. ok, i'll tweak the regex

jackocnr:
Great thanks. Yeah I feel like there must be some nice generalised regex to solve this for all languages, but don't have time to look into it myself right now, so I'm happy to just solve your problem for now 👍

PR #1034, opened 2020-02-17
Fix URL to type definitions

toddsiegel:
This was going to an old, broken URL.

The updated URL takes you to the type definition file.


jackocnr:
Nice one

PR #1037, opened 2020-02-24
added option updateHiddenOnChange

ionutab:
Added a boolean flag to the plugin, updateHiddenOnChange.
I added this flag because the hidden input was not updated automatically when I was handling the form via a javascript submit.
So, with this flag, the hidden input will be updated onChange rather than on form submit.


jackocnr:
Hi there, thanks for your work on this, but I'd like to discuss it a bit before I consider merging it, in order to decide if it's the right direction to go in.

So basically I'd really like to avoid these kinds of fine-grained options - I'd really like to keep the plugin as simple and lightweight as possible. But maybe this is just a better approach than the submit listener, and this should just be the default? What do you think? Does this cover all of the same cases as the submit handler, and more? What is browser support like?

Cheers.

ionutab:
Hello,

**I appreciate your idea of keeping things simple, as they should be, and indeed this is a fine-grained option.**

However, I'm not sure how many people use the submit handler on forms. For example, consider if the form holding the telephone number input is in a modal that has to close on submit and the form is submitted via ajax.

When initializing with `hiddenInput: "full"` then the hidden input should have the full telephone number upon submit. I believe it is preferable that this works on both on ajax and non-ajax submitted forms in order to avoid confusion for developers.

What happened to me was that I started adding the plugin to my project, tested it on two static pages and then proceeded to add it in some modals as well, and that is when I found this difference of behaviour. I'm not sure how people submit modal forms that don't redirect to other pages other than via some javascript code.

I am not a proficient javascript developer.
I want to say it covers all of the cases, since I've tested this in my project and I have not had any issues, however, I would not trust myself 100% based on my experience writing js.
I have no idea how to test browser support.

To be honest I think this whole side of the plugin depends on if the user wants the full telephone number or not. Because if the user wants to send the non-prefixed telephone number then things should stay as they are. When the user wants the full telephone number there is this hidden magic we have to do in order to send it, and if the hidden input is the best option I see no reason why the initial input and the hidden one should not always be in sync.

-----
Now, there's another thing I want to add.
It's something I did in my project.
Since I had about 30 tel number inputs across different pages/modals, with js listeners and all and I did not want to change the input names but wanted to get the full telephone number on the server side upon submit, I did something like this:
```
skilloIntlTelInput: function (options) {
            let defaults = {
                preferredCountries: ['gb', 'ie', 'ro', 'nl', 'bg', 'pl', 'rs'],
                utilsScript: '/skillo-plugins/intl-tel-input/js/utils.js',
                updateHiddenOnChange: true
            };

            this.each(function () {

                const myTelInput = $(this);
                const telephoneNumberInputName = myTelInput.attr('name');
                myTelInput.attr('name', 't_' + telephoneNumberInputName);

                if (options !== undefined) {
                    myTelInput.intlTelInput(options);
                } else {
                    Object.assign(defaults, {
                        hiddenInput: telephoneNumberInputName
                    });
                    myTelInput.intlTelInput(defaults);
                }
            });
            return this;
        }
```
This changes the name of the initial telephone number input but moves the initial name to the hidden input. In this way I don't have to have a diferent name for each hiddenInput in my project.
This second part is just something I did in order to make the integration as fast as possible without changing the html/js code too much.

jackocnr:
> However, I'm not sure how many people use the submit handler on forms. For example, consider if the form holding the telephone number input is in a modal that has to close on submit and the form is submitted via ajax.

If the form is submitted via Ajax then you don't need a hidden input. You are in control of the request being made, and the data sent, so you can just call `getNumber` and pass the result in your request.

ionutab:
Ok, 
this would solve the issue, indeed.
Should I add a section in the documentation stating this, with a short example as well?

jackocnr:
You could add a sentence to the `hiddenInput` description if you like. We already say:

> This is a quick way for people using non-ajax forms to get the full international number

So maybe after that sentence you could add a sentence about how this option is not needed for ajax forms, where you can just call getNumber yourself, and pass the result in the request.

Thanks. I'll close this PR for now.

ionutab:
Ok. Perfect. I will add this sentence and make another pull request.
Thank you.

Issue #707, opened 2018-03-09
window is not defined issue


niangoss:
IMPORTANT: please note that the autoFormat feature (format-as-you-type) was removed in v8.0.0 - read more here: https://github.com/jackocnr/intl-tel-input/issues/346. Please do not open any new issues about this topic.

IMPORTANT: we use libphonenumber for formatting and validation - if your issue relates to one of these things, please check their test site first and if you see the problem there please file an issue with them instead: https://rawgit.com/googlei18n/libphonenumber/master/javascript/i18n/phonenumbers/demo-compiled.html

### Steps to reproduce
When you use it in angular 5 universal server side rendering project. 
window exist only on browser side, we got  "window is not defined" issue when factory($, window, document) is called (build/js/intlTelInput.js file at line 11)
 

### Expected behaviour
check if 'window' exists before use it

### Actual behaviour
window is not defined Exception is thrown when window does not exist

### Initialisation options
List any options you're using e.g. utilsScript or preferredCountries

jackocnr:
Hmm interesting! I would be open to a pull request to check if window exists before attempting to use it, and if not, then it immediately sets `windowLoaded` to true. I just checked and we reference `window` quite a few times in the code, but I suppose only the initialisation path is relevant for server side rendering.

yacineblr:
Do you have any news?

jackocnr:
No update yet I'm afraid. But a quick search revealed this thread on the angular universal repo: https://github.com/angular/universal/issues/830 which suggests that when you have a problem with a module trying to access `window` - that you either (A) only import that module in main.ts (which I'm guessing means it wont be run on the server side?), or (B) define `global['window']` on your backend. So I'm not convinced we actually need to do anything about this in the intl-tel-input plugin code. What are your thoughts?

I have never used Angular, or Angular Universal, but I'm interested to know: do you actually want to run this plugin code on the server side? And if so, why?

jackocnr:
Closing due to inactivity

PR #1057, opened 2020-03-20
Fix “ReferenceError: window is not defined” when importing code in Node.js

jacksonrayhamilton:
I had a problem similar to #707.

I have some JS files that are glued together with Webpack, which looks at the ES6 `import` syntax I use to determine which files to bundle up.  This means I am loading intl-tel-input in both my Node.js server code and my browser code.  My Node.js server code is going to render some HTML using another library but skip calling `intlTelInput()` on the server.  Then when the code is loaded in the browser `intlTelInput()` will enhance my input.

The problem is that if I so much as add `import intlTelInput from 'intl-tel-input'` to my server code, I get an error: `ReferenceError: window is not defined`.

I could dynamically import my dependency with `await import('intl-tel-input')` to avoid the ReferenceError, but that adds complexity.  It would be much nicer if all I needed to do was `if (typeof window === 'object') intlTelInput(domNode)`.

I think that with a slight modification to the style in which we initialize the library, such that we write in a more environment-agnostic [(“isomorphic”)](https://cdb.reacttraining.com/universal-javascript-4761051b7ae9) style, this use case can be served.  I believe that f210096 achieved this.

I made the change 8301a94 in response to what you said about `windowLoaded` in this comment: https://github.com/jackocnr/intl-tel-input/issues/707#issuecomment-371783661

~I went ahead and extended the code style further in 5fbe95c to be more consistent.  If you are amenable to this broader change then the tests could probably benefit from it too.  But in case you don’t like that, we can roll it back.~  _(Whoops, that broke the tests.  Nevermind.)_

Thanks for taking your time to review this.


jackocnr:
Thanks so much for this @jacksonrayhamilton, I think the "Avoid referencing..." commit is perfect and ready to merge.

As for the "Synchronously..." commit, while I appreciate that I did say that in that old thread, in this context (just importing the plugin, but not initialising) there's no point as that value wont be read at all. I think if we wanted to look into allowing you to initialise the plugin on the server, then we'd need to consider it then, but in that case there's a lot more to consider that we haven't yet. So can we remove this for now?

And finally, in the "Build changes..." commit, it seems to be injecting a typeof polyfill - any idea what is causing this to be added and why? I don't think we need this do we?

Thanks again, I look forward to getting this merged.

jacksonrayhamilton:
> As for the "Synchronously..." commit … So can we remove this for now?

Sure, I removed it.  I also rebased on v16.0.15 and force-pushed to this branch.

> And finally, in the "Build changes..." commit, it seems to be injecting a typeof polyfill - any idea what is causing this to be added and why? I don't think we need this do we?

I imagine that Babel adds that to be extra extra safe.  ES6 added `Symbol` and extended the range of possible return values of `typeof`.  Theoretically, the code could run in ES5 _and_ the environment could have a `Symbol` polyfill _and_ the environment could assign `window = Symbol('window')` _but_ the `typeof window` operations in intlTelInput.js would incorrectly return `"object"` rather than `"symbol"` in that environment.

Highly unlikely, I know.  But is it worth configuring Babel to not do it?  Eh, I prefer to generally assume that preset-env is wiser than I am about compiling JS.  And it’s only a few extra bytes.

jackocnr:
> Sure, I removed it. I also rebased on v16.0.15 and force-pushed to this branch.

Amazing, thanks.

> I prefer to generally assume that preset-env is wiser than I am about compiling JS. And it’s only a few extra bytes.

I totally agree in the general case, but in the case of a plugin which will be imported into other people's projects - I think it's their job to decide the Babel config that applies to all of their code.

jacksonrayhamilton:
Alright, I disabled the `typeof` transform.

jackocnr:
Thanks so much for this. And I think excluding the typeof transform was the right thing to do - this article discusses it, and how it's only needed if you're using typeof with Symbols: https://medium.com/@WebReflection/avoiding-babels-production-bloat-d53eea2e1cbf

PR #1064, opened 2020-04-02
Update libphonenumber to v8.12.1

dking-tyro:


jackocnr:
Thank you so much for putting this together, but in this case the diff ended up being way too complicated for me to review safely, so I'll just do this one myself. Thanks again.

PR #1066, opened 2020-04-06
Feature: autoComplete=polite

luyseyal:
If you like this approach to autoComplete=polite, I can push up some documentation and a test for it.


jackocnr:
Thank you so much for jumping on this.

Can we can do this without a new initialisation option? Can we literally just check if there is already an autocomplete attribute on the input, and only add `autocomplete=off` if not?

luyseyal:
@jackocnr I was worried that intl-tel-input might be used as a bolt-on to a framework which defaults to adding autocompletes and you might end up with mysterious "why doesn't the flag match in Drupal/Sugar/whatever?" bugs... especially when it might depend on browser load order.

Is that valid? Or overthinking it? I don't have an example or anything...

jackocnr:
Hmm it's a good thought, but it's not an issue I've heard of yet, and I'd really rather avoid adding any more initialisation options until we really have to. Thanks.

luyseyal:
Updated PR. Checking element and its form for extant autocomplete before setting it to off.

jackocnr:
Manually merged, thank you! This will be in the next release.

PR #1084, opened 2020-05-05
Update data.js to include Christmas Island and Cocos Island areacodes

dgrochmal:
https://en.wikipedia.org/wiki/Telephone_numbers_in_Christmas_Island
https://en.wikipedia.org/wiki/Telephone_numbers_in_the_Cocos_(Keeling)_Islands
https://en.wikipedia.org/wiki/List_of_country_calling_codes

### DIFF
--- a/src/js/data.js
+++ b/src/js/data.js
@@ -260,13 +260,15 @@ var allCountries = [
     "Christmas Island",
     "cx",
     "61",
-    2
+    2,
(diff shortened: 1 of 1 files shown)


jackocnr:
This looks like a good change, thanks. I just need to find some time to do some smoke testing before merging.

jackocnr:
Thank you

PR #1094, opened 2020-05-20
Format and removing last line that is incorrect

elminson:


jackocnr:
You'll notice that file is called "outro-jquery" - if you look at "intro-jquery" you'll see why we need those extra closing brackets (see build/js/intlTelInput-jquery.js for the finished product)

PR #1100, opened 2020-05-29
Fix issue 1089

tomas-nz:
Don't replace area codes of NANP numbers when changing country code.

### DIFF
--- a/src/js/intlTelInput.js
+++ b/src/js/intlTelInput.js
@@ -1119,7 +1119,7 @@ class Iti {
     let newNumber;
     if (inputVal.charAt(0) === '+') {
       // there's a plus so we're dealing with a replacement (doesn't matter if nationalMode or not)
-      const prevDialCode = this._getDialCode(inputVal);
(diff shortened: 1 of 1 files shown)


jackocnr:
Thanks for this. It's saying there are conflicts - would you be able to resolve those? And ideally squash commits into a single one.

tomas-nz:
@jackocnr I managed to squash the commits. The message above here on the PR says "This branch has _no conflicts_ with the base branch" and has a green tick. If I missed something, kindly point me in the right direction.

jackocnr:
That seems to have fixed the conflict issue, thanks.

jackocnr:
Thanks for this PR. I agree with your conclusion that it's best not to touch any area codes when changing flags, for consistency.

Unfortunately this solution only fixed the issue for NANP countries, but not for other countries where we handle dial codes e.g. +61. I've now pushed a more comprehensive fix to master, so I'll close this. Thanks for your help driving this forwards.

PR #1106, opened 2020-06-08
Allow for dynamic component importing (after page load)

pauliclark:
The issue was discovered while including in a stencil component. The utilsScript was not loading when the parent component was dynamically loaded after page load. Therefore, I have added a check for the document ready state.
Also... upgraded vulnerable packages


jackocnr:
This change looks good, thank you, I just need to find some time to smoke test - apologies for the delay.

alabbas-ali:
Are you planing on merging this changes at some point ???

jackocnr:
Thanks so much for this. I decided to drop the package updates for now - I think we should handle those separately (and do lots of testing that all of the build processes still work properly). I manually merged your commit, so I'll close this PR, thanks again. This will be included in a new release later today.

PR #1108, opened 2020-06-17
Update CSS file location

akshayp660:
by adding style and its location, the file will let user import intlTelInput.css to project via webpack or laravel-mix.


jackocnr:
Thanks

PR #1112, opened 2020-06-27
Fix: bug in validation short number error message

boradakash:
It's showing undefined as error message in validation when only single digit is entered. That's fixed in this commit.

Bug:
image


jackocnr:
Looks good, thank you

PR #1118, opened 2020-07-10
webp format for image compression 

akshayp660:
added .webp format flag images to support fast loading without loosing image quality, please re-compile assets


jackocnr:
Hmm but what about browsers that don't support webp?

akshayp660:
in that case you must add fallback code to load .png format what you think

akshayp660:
committed code that will help to load .png format  if browser does not support .webp please check .scss file

jackocnr:
Thanks for this, but there are a couple of problems: (1) the image sprite is generated automatically, and subject to change, so if we were going to add support for this, we would need to hook it into that build system, but also (2) I think that handling modern browser-specific image formats is up to the dev to do their own way - for instance, recently I've been using Thumbor on the server, which automatically converts any image to the best format that is supported by the browser that requested it.

So thanks for the idea, but I don't want to add this responsibility into the plugin right now - I'd rather leave it up to the dev.

PR #1144, opened 2020-09-11
Update ARIA tags

iesoftwaredeveloper:
Add aria-label to meet WCAG 2 A accessibility


jackocnr:
Thanks so much. This will be included in a new release later today.

Issue #1158, opened 2020-10-12
Uncaught SyntaxError: invalid range in character class in the Firefox


wppunk:
I've got the error in Firefox for developers in this regular extension `/^[a-zA-ZÀ-ÿа-яА-Я ]$/`.
`Uncaught SyntaxError: invalid range in character class` in `ÿ`.

### Steps to reproduce
1. Include scripts for intl-tel-input for jquery
2. Open in Firefox for developers 

### Initialisation options
utilsScript

wppunk:
`À-ÿ` it's the [incorrect range](https://sites.psu.edu/symbolcodes/languages/psu/french/) for the French letters because these letters are not range.
The correct way it's just `àâäèéêëîïôœùûüÿçÀÂÄÈÉÊËÎÏÔŒÙÛÜŸÇ`

jackocnr:
I just installed Firefox Developer edition and visited [the demo site](https://intl-tel-input.com) and didn't see any errors in the JS console. I even typed out the regex to try it in the console and it worked fine:

<img width="249" alt="Screenshot 2020-10-20 at 09 32 08" src="https://user-images.githubusercontent.com/1186883/96561300-48862c80-12b7-11eb-8005-2aea972a388e.png">

Please can you give clear steps to reproduce the error you're seeing?

wppunk:
@jackocnr Oh sorry, I can't reproduce this bug right now. But I have another argument for these changes: https://regex101.com/r/3whJIY/1

jackocnr:
Right, but if you enter the actual full regex that we use then it matches a lot more of those chars. In fact it seems to match all of them except 3.

As for those last 3 - I'm afraid I'm not familiar with them, but if you think they're important, then if you can point me to a reference for why we should include those last 3 chars, then I'll happily accept a PR to add them to the regex :)

jackocnr:
Closing due to inactivity.

mtarregh:
I'm having the same issue with 17.0.12. @wppunk were you able to find a workaround?

wppunk:
@mtarregh I guess that it was FireFox bug

alexandrepa:
I had the same issue in Firefox with 17.0.12
Adding `<meta charset="utf-8">` in the `<head>` of my html file solved the problem for me.

mtarregh:
For me it actually turned out to be a missing default-encoding in the JBoss config that lead to the script not being encoded correctly.

aabid-t:
[...]

PR #1159, opened 2020-10-12
Uncaught SyntaxError: invalid range in character class in the Firefox

wppunk:
Fixes: #1158

`À-ÿ` it's the incorrect range for the French letters because these letters are not range.
The correct way it's just `àâäèéêëîïôœùûüÿçÀÂÄÈÉÊËÎÏÔŒÙÛÜŸÇ`.


jackocnr:
Hi there, thanks for this. Do you have a reference as to why the current regex is invalid and this new regex is the correct way?

Also there seems to be a conflict - please make sure you're up to date with master, thanks.

wppunk:
When I've used the library in the Firefox then I can see the error message `Uncaught SyntaxError: invalid range in character class` It's mean what this range it's incorrect.

Conflict with the master in the minified build scripts.

PR #1162, opened 2020-10-15
Add build utils script

NamPNQ:
Add a script for build `utils.js`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "third_party/libphonenumber"]
+	path = third_party/libphonenumber
+	url = https://github.com/google/libphonenumber
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -3,12 +3,17 @@ module.exports = function(grunt) {
   // load all tasks from package.json
   require('load-grunt-config')(grunt);
   require('time-grunt')(grunt);
+  require('google-closure-compiler').grunt(grunt, {
+    platfrom: 'native'
+  })
 
   /**
    * TASKS
    */
   // build everything ready for a commit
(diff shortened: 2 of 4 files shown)


jackocnr:
Thanks so much for submitting this - how exciting! I just need to find the time to understand and smoke test this properly, so please bear with me.

jackocnr:
@NamPNQ believe it or not, I am finally getting around to looking into this! I can only apologise for the ridiculous delay, but I hope you understand that open source work can suck up a lot of hours of your life for very little to no money, hence it being de-prioritised in recent years. Now I'm assuming you have likely moved on with your life 😂  but on the off chance you're still interested in this project, I'll share my thoughts: this would absolutely make a brilliant addition to the project, and I have gotten it building on my end, with lots of warnings but no errors! The only problem is that (A) when I build a new utils.js file, the file doesn't seem to work for me, despite it building without errors, and the contents of the file even looking similar to how they did before, unfortunately when I load it in the demo, I don't get any errors, but the placeholder number never appears which means it never initialises properly. And (B) the filesize of utils.js jumps up from 250kb to 330kb (that's with me disabling debug mode in grunt/closure-compiler.js). Do let me know your thoughts if you feel like resurrecting this dinosaur! But of course, I completely understand if this project no longer interests you.

NamPNQ:
Hello @jackocnr, hope you're doing well

I just updated the build script and all test is passed :)

jackocnr:
Amazing, this is working perfectly, thank you so much @NamPNQ! The built utils.js works perfectly, tests pass as you say, and the build file is a similar size to before. Thanks again!

PR #1168, opened 2020-11-03
Allow multiple instances on a page

timbreitkreutz:
We are using this plug-in to add forms to our customers' web sites, via a packaged script we have them include on their page. This can interfere with our customers' instances of the `intl-tel-input` plugin that they have used for other purposes.  This is caused by our copy overwriting the global object in `window.intlTelInputGlobals`.

Currently we work around this by using a local modified copy of this repo, but we would prefer to use the regular distribution of this plugin.

This PR suggests a way to avoid clobbering that global by checking to see if there is one already, then iterating with increasing number suffixes (i.e. `intlTelInputGlobals1`, `intlTelInputGlobals2`, etc.) until it finds one that is available.

## Alternative ideas

After trying a few different approaches, the one in this PR seemed to be the least invasive option. However, there could be other advantages to the following alternatives:

* Instead of using a suffix number, an array could be created inside the main `intlTelInputGlobals` to contain nested instantiation data.
* a version of the script could be added to the build (via a different grunt header and footer) that doesn't require or expose the global object and/or allows a global object name to be passed in from the client after includes.


timbreitkreutz:
Another question:  This only takes care of clobbering the Globals object itself, not the utils.  I think the utils are probably less likely to be a problem, but if you like we could try to follow the same pattern for that if it makes sense.

timbreitkreutz:
@jackocnr do you think this might be considered for a release?  We'd prefer not to use our own version if possible.  Thanks! :)

timbreitkreutz:
Update: We have been running this version in our main app for a couple weeks now and have no reported issues.

timbreitkreutz:
ping @jackocnr  it would be cool if this was an OK change for you so we could revert to the public version... thanks :).

timbreitkreutz:
@jackocnr would it be worthwhile for me to update this PR to the new stuff on the main branch?

jackocnr:
Hi there, apologies for my silence on this, I've been very busy.

I'm afraid I don't want to add all of this complexity to the core lib, as I think this kind of use will be incredibly rare.

With that said, I would be open to a lighter touch solution if we can think of one. For instance, I wonder if we could just avoid overriding the global, and just have all instances share the same one? Do you think that could work?

timbreitkreutz:
> Hi there, apologies for my silence on this, I've been very busy.
> 
> I'm afraid I don't want to add all of this complexity to the core lib, as I think this kind of use will be incredibly rare.
> 
> With that said, I would be open to a lighter touch solution if we can think of one. For instance, I wonder if we could just avoid overriding the global, and just have all instances share the same one? Do you think that could work?

Yes I think that would work and I did think of doing it that way--the risk there would be two different versions sharing the same global.  IIRC (it's been a few months) it's really only the instances in the global that are a problem so those could also be stored in a safer manner.

timbreitkreutz:
I'll see if I can find some time to revisit one of the alternative approaches discussed above.  I think this PR was actually the cleanest approach and it would have been an improvement to the codebase.

PR #1175, opened 2020-11-17
Add accessibility attributes for selected dial-code node

ElenaShevchenko:
According to accessibility requirements "Authors MUST ensure an element with role combobox contains or owns a text input element with role textbox or searchbox", so I add texbox role to country code element + appropriate aria attributes
https://www.w3.org/TR/wai-aria-1.1/#combobox

### DIFF
--- a/src/js/intlTelInput.js
+++ b/src/js/intlTelInput.js
@@ -339,7 +339,12 @@ class Iti {
     this.selectedFlagInner = this._createEl('div', { class: 'iti__flag' }, this.selectedFlag);
 
     if (this.options.separateDialCode) {
-      this.selectedDialCode = this._createEl('div', { class: 'iti__selected-dial-code' }, this.selectedFlag);
+      this.selectedDialCode = this._createEl('div', {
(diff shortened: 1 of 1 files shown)


ElenaShevchenko:
According to accessibility requirements "Authors MUST ensure an element with role combobox contains or owns a text input element with role textbox or searchbox", so I add texbox role to country code element + appropriate aria attributes 
https://www.w3.org/TR/wai-aria-1.1/#combobox

jackocnr:
Hi there, apologies for the delay in responding. Thank you for this suggestion, but after reading a bit about this, I don't think we need it. Looking at this [W3C Aria Combobox example](https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html), which I think is the same use case as the dropdown in this plugin, they don't have this role=textbox thing. As fart as I can tell, this role=textbox is only for when you display a text input for searching within the list of options, which this plugin does not have. I'll close this for now, but do let me know if you think I've misunderstood, thanks.

PR #1192, opened 2021-02-09
_getFullNumber trim spaces 

Tusko:
I've fixed the method `_getFullNumber` to replace all special characters (for example if I use the mask plugin) to validate the full number

### DIFF
--- a/src/js/intlTelInput.js
+++ b/src/js/intlTelInput.js
@@ -1196,7 +1196,7 @@ class Iti {
 
   // get the input val, adding the dial code if separateDialCode is enabled
   _getFullNumber() {
-    const val = this.telInput.value.trim();
+    const val = this.telInput.value.replace(/[^0-9+]+/g, "");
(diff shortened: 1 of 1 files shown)


aguchux:
Nice one brother.

jackocnr:
My apologies for the delay in replying - I've been dealing with a family emergency.

I'm afraid I don't understand this change. As far as I can see, you've changed the `_getFullNumber` private method so that it drops any non-numeric chars. But this method is only invoked in 5 places, and in every case, the returned value just gets passed to one of a few methods in utils.js, all of which interally use libphonenumber's `parseAndKeepRawInput`, which should ignore any non-numeric chars anyway. Am I missing something?

With all things like this, it's best to start by opening an issue to discuss the problem you're having - give me steps to reproduce the problem vs what you would expect to happen, and then together we can decide if it's something we need to fix, and if so, what's the best way.

Tusko:
It was 1 year ago.
I do not remember. You can remove the PR

Issue #1198, opened 2021-02-15
Cannot update libphonenumber lib.


OsaAjani:
Hello, i wanted to update libphonenumber lib in order to get the latest change concerning the recent format change of ivory coast phone numbers, though i cannot update the lib following the guide in CONTRIBUTING.

I've been searching for a while, but i'm no expert with JS and Java compiling tools and i struggle to find a solution.

### Steps to reproduce
1.  Follow the guide https://github.com/jackocnr/intl-tel-input/blob/master/.github/CONTRIBUTING.md#updating-to-a-new-version-of-libphonenumber

### Expected behaviour
We should be able to update the libephonenumber lib.

### Actual behaviour
I got a few different errors :
1. In step 1, the guide say to run `mvn -DskipTests` in closure-compiler directory. But we got the error "No goals have been specified for this build.". More generally it seems closure-compiler now use Bazel instead of Maven. Anyway, i found a way to compile closure-compiler by following the "Building from a terminal" part from the README of the closure-compiler.
2. In step 2, when i run the command `ant -f libphonenumber/javascript/build.xml compile-demo` i got this error. 
```Buildfile: /home/ajani/workspace/libphonenumber-tools/libphonenumber/javascript/build.xml

compile-demo:
     [exec] /home/ajani/workspace/libphonenumber-tools/libphonenumber/javascript/../../closure-library/closure/bin/build/closurebuilder.py: Scanning paths...
     [exec] /home/ajani/workspace/libphonenumber-tools/libphonenumber/javascript/../../closure-library/closure/bin/build/closurebuilder.py: 1656 sources scanned.
     [exec] /home/ajani/workspace/libphonenumber-tools/libphonenumber/javascript/../../closure-library/closure/bin/build/closurebuilder.py: Building dependency tree..
     [exec] /home/ajani/workspace/libphonenumber-tools/libphonenumber/javascript/../../closure-library/closure/bin/build/closurebuilder.py: No namespaces found. At least one namespace must be specified with the --namespace or --input flags.

BUILD FAILED
/home/ajani/workspace/libphonenumber-tools/libphonenumber/javascript/build.xml:99: The following error occurred while executing this line:
/home/ajani/workspace/libphonenumber-tools/libphonenumber/javascript/build.xml:23: exec returned: 2

Total time: 1 second
```

OsaAjani:
To follow up, i've managed to update libphonenumber but it was quite a challenge and the CONTRIBUTING file really need to be updated.

[...]

Issue #1189, opened 2021-02-04
The new format of the ivory coast telephone number is not supported


chaabenMohamed:
Ivory coast changed its telephone number format on 31 January 2021, from 8 to 10 digits.
https://news.abidjan.net/h/677162.html

### Steps to reproduce
1.  try to whrite a phone number for ivory coast with 10 digits

### Expected behaviour
The phone number should be valid

### Actual behaviour
The phone number is not accepted as valid phone number for Ivory coast

jackocnr:
We use [libphonenumber](https://github.com/googlei18n/libphonenumber) for formatting, validation and generating placeholder numbers - if your issue relates to one of these things, please check [their test site](https://rawgit.com/googlei18n/libphonenumber/master/javascript/i18n/phonenumbers/demo-compiled.html) first and if you see the problem there please file an issue with them instead.

If it's working correctly in libphonenumber, the next step is to check that we're using [the latest version](https://github.com/google/libphonenumber/releases) of libphonenumber in the plugin - check the version history [here](https://github.com/jackocnr/intl-tel-input/commits/master/build/js/utils.js). If we need to upgrade, it only takes 5 minutes to do this yourself and submit a Pull Request - instructions here: [Updating to a new version of libphonenumber](https://github.com/jackocnr/intl-tel-input/blob/master/.github/CONTRIBUTING.md#updating-to-a-new-version-of-libphonenumber).

jackocnr:
Note: we just released a new version with an upgraded libphonenumber - please can you see if that fixes your issue?

chaabenMohamed:
The issue doesn't exist in libphonenumber, their demo page: https://libphonenumber.appspot.com/phonenumberparser?number=0748436784&country=CI
But the bug still exists in intl-tel-input in the last release!

jackocnr:
Hmm I see libphonenumber have actually just released a new version v8.12.18 (5 hours ago!), which does include updates to Ivory Coast meta data - if we update to that it should solve the issue.

jonathan42:
#1213 ?

jackocnr:
I've just published a new version (v17.0.13) using a newer version of libphonenumber, which should have fixed the issue. Feel free to post here if you're still seeing the issue tho.

PR #1200, opened 2021-02-20
Update libphonenumber to v8.12.18

OsaAjani:
This pull request update src/utils.js to use libphonenumber v8.12.18 (including among other things transition from 8 to 10 chars phone numbers for Ivory Coast #1189). #1198


Bartlebyy:
While the minified version might be faster, it also makes that file completely unreadable and uneditable going forward. I'd rather the compiler minify the file and leave this file readable

Bartlebyy (changes requested):
@OsaAjani While the minified version might be faster, it also makes that file completely unreadable and uneditable going forward. I'd rather the compiler minify the file and leave this file readable

OsaAjani:
Absolutly, my mistake i havent realized the file have been minified in the building process. I will have to find where in the process the file is minified and how to disable it.

OsaAjani:
My bad, i've been copying the minified version in `src` instead of `build`. So i basicly copied back the old utils.js in src, add the namespace without with i cannot compile the script, and add the minified version to build dir.

jonathan42:
#1213

jackocnr:
Thanks so much for putting this together, and apologies for my delay in reviewing - I've been dealing with a family emergency. I've just bumped us up to v8.12.21, so I'll close this one now. Thanks again.

PR #1208, opened 2021-03-10
Add the namespace and require to fix the build with the recent Google Closure compiler

cmrd-senya:
I tried building the utils with the recent Google Closure compiler following this instructions: https://github.com/jackocnr/intl-tel-input/blob/master/.github/CONTRIBUTING.md#updating-to-a-new-version-of-libphonenumber

I faced the same issue that is mentioned here https://github.com/jackocnr/intl-tel-input/issues/1198

I've managed to fix the issue by updating both code at the libphonenumber side and intl-tel-input side. This PR is the changes required to be able to build `utils.js` with the recent Google Closure compiler version.

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -1,6 +1,8 @@
+goog.provide('i18n.phonenumbers.demo');
 // includes
 goog.require('i18n.phonenumbers.PhoneNumberFormat');
 goog.require('i18n.phonenumbers.PhoneNumberUtil');
+goog.require('i18n.phonenumbers.Error');
 
 
 // format the given number to the given format


jackocnr:
Thank you very much for this necessary fix 🙏 

Unfortunately, with this new (required) "demo" import, when building the latest version of libphonenumber (v8.12.24 at the time of writing), we're adding 50kb to the built utils.js file, raising it from the already massive ~250kb to ~300kb 😞

jackocnr:
Hmm @cmrd-senya in addition to the concern about filesize, the new build (based on libphonenumber 8.12.24) also breaks the Jasmine tests, spitting out CORS errors for some reason - perhaps there's some error in the new utils build file or something? Did you try it out? Unfortunately I don't have time to look into this, due to a family emergency. Let me know if you have any time to check this out. 🙏

cmrd-senya:
@jackocnr thanks for the mention, I'll take a look. Last time when I was testing this the Jasmine tests were running fine

PR #1213, opened 2021-03-17
Update libphonenumber to v8.12.20

jonathan42:


jackocnr:
Thanks so much for putting this together, and apologies for my delay in reviewing - I've been dealing with a family emergency. I've just bumped us up to v8.12.21, so I'll close this one now. Thanks again.

PR #1223, opened 2021-04-28
Add a data value to flag with country prefix

tadywankenobi:
Found that required the actual country code prefix for working with 3D Secure payments with Global Payments. Needed a way to have just the country code prefix so set it to a data value (data-intl-prefix) in the flag element.


tadywankenobi:
Should have added, was just suggesting this as a feature update. Not necessary, just thought it might be useful to others.

jackocnr:
Hi there, you can get this info using the `getSelectedCountryData` method.

PR #1237, opened 2021-06-03
Update libphonenumber version to v8.12.20

Alweezy:
* Update libphonenumber version to access `metadata changes`.


jackocnr:
Thank you for putting this together. Unfortunately [the tests are failing](https://travis-ci.org/github/jackocnr/intl-tel-input/jobs/773384671) - lots of "Content Security Policy" errors saying the test scripts could not be loaded into the page.

> log: Refused to load the script 'file:///home/travis/build/jackocnr/intl-tel-input/src/spec/tests/core/closecountrydropdownEvent.js' because it violates the following Content Security Policy directive: "script-src 'nonce-CSP+Nonce+For+Tests+Only' 'unsafe-inline' 'unsafe-eval' 'strict-dynamic' https: http:". 'strict-dynamic' is present, so host-based whitelisting is disabled. Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.

Did you smoke test this new utils.js file? Was it working ok for you? (easiest way to test it locally is to just open the root demo.html file in your browser).

cc: @cmrd-senya

Alweezy:
Hi @jackocnr I'm actively looking through to see what could be the isssue.
I guess I went up a huge leap of versions.
I have just tested a lower version `v8.12.20` and that is working locally.
unfortunately I'm unable to trigger travis builds.
Probably because I force pushed.
Do you know how I can retrigger the builds?

Alweezy:
Hi @jackocnr , sorry about that.
It seems the later versions of `libphonenumber` have some changes with our package.
I have downgraded to `v8.12.20` and that works.
We howver have this error on `travis-ci ` that npm install failed, I'm not sure what the cause is, I think you experienced it with `17.0.13` too.

jackocnr:
The latest version of the plugin is built against libphonenumber v8.12.21, so no need for this.

Do let me know though if you're able to test the build with the latest version of libphonenumber.

kissifrot:
Do you have news regarding the version update? 

We're getting phone number validation errors with v8.12.21 (SG number 8939xxxx), but it's fixed with upstream version v8.12.24+

Awkan:
Hello and thanks for this PR :pray: 

Waiting for news too ... I need at least `libphonenumber^8.12.22` (best upgrade would be last release : v8.12.26 released on June 25, 2021)

chloe-1903:
Same here, the update would be nice since they fixed the Philippine number validation ([opened issue](https://github.com/jackocnr/intl-tel-input/issues/1242))

jackocnr:
I've just released a new version (v17.0.14) which is built against the latest version of libphonenumber. Thanks.

PR #1239, opened 2021-06-07
Update North Macedonia country name

beingke:
Based on recent name dispute resolution: https://en.wikipedia.org/wiki/Macedonia_naming_dispute

The PR doesn't include the output of ```npm run build:js```.


jackocnr:
Thanks

Issue #1241, opened 2021-06-15
Disable keyboard navigation on a disabled input


loicplaire:
### Steps to reproduce
1.  Open https://codesandbox.io/s/relaxed-ptolemy-vopdk
2.  Using the mouse, set the focus on the codesanbox preview area.
3.  Use the keyboard `tab` key to set the focus on the country dropdown.
4. Press the `arrow down` key.
5. The country dropdown opens and you can select a country 😕.

![intl-tel-input](https://user-images.githubusercontent.com/1925075/122136141-3465ef80-ce85-11eb-9d36-af9baa034ba2.gif)

### Expected behaviour

Using keyboard navigation, it shouldn't be possible to focus and open the country dropdown.

### Actual behaviour

When applied to a `disabled` input, it is still possible to open the country dropdown and select a country using the keyboard.

### Potential solution
Setting the `tab-index` to `-1` when the input is disabled would fix the issue. We should also look at adding the `aria-disabled` attributes for a11y purpose.

### Initialisation options

```js
intlTelInput(inputEl.current, {
  utilsScript:
    "https://cdn.jsdelivr.net/npm/intl-tel-input@17.0.3/build/js/utils.js"
});
```

jackocnr:
Good catch. I would welcome a PR to address this. I wasn't sure about the idea of making the selected flag un-tabbable tho - I guess it should behave the same as normal inputs. I don't have time to check myself right now, but can you tab to normal inputs when they're disabled?

loicplaire:
Thanks @jackocnr, I should be able to take a look at it in the next few days.
An element with the `disabled` attribute is not focusable: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled so as you said, the same logic should apply to the dropdown.

PR #1244, opened 2021-06-21
Prevent dropdown focus when the associated input is disabled

loicplaire:
# Explanation

See the related issue for more info but in short, you shouldn't able to focus on the dropdown using the keyboard when the associated input is disabled.

**Related issue:** https://github.com/jackocnr/intl-tel-input/issues/1241

# Changes

- Set the `tabindex` to `-1` and `aria-disabled` to `true`  when the associated input has the `disabled` attribute.
- Added tests.
- Ran `npm run build:js`


src/spec/tests/core/usingDropdown.js
```diff
@@ -79,3 +71,25 @@ describe("using dropdown: init plugin with nationalMode=false", function() {
...
+    iti = window.intlTelInput(input[0]);
+  });
+
+  afterEach(function() {
+    intlTeardown();
+  });
+
+  it("prevents the user from opening the dropdown using the keyboard", function() {
+    expect(getSelectedFlagContainer().attr('tabindex')).toEqual('-1');
+    expect(getSelectedFlagContainer().attr('aria-disabled')).toEqual('true');
```
loicplaire, line 88: FYI, I tried the following:

```js
getSelectedFlagContainer()[0].dispatchEvent(new CustomEvent('focus'));
triggerKeyOnFlagsContainerElement("ArrowDown");
expect(getListElement()).not.toBeVisible();
```
But this doesn't work because `dispatchEvent` doesn't check the `disabled` attribute:  https://github.com/testing-library/dom-testing-library/issues/92#issuecomment-530249041

VanonckelenDieter:
What is the status of this PR? Good catch by the way!

jackocnr:
Apologies. I will have some time in April to catch up on these PRs.

jackocnr:
Thank you so much for this, it looks great. Can I ask for a slight change before we merge:

- Can we just check the value of `this.telInput.disabled` directly, which is something we do elsewhere in the code
- Can we move the `aria-disabled` bit up to straight after the `this.selectedFlag` definition e.g. `if (this.telInput.disabled) [add attribute]`, as I don't think this should be dependent on the `allowDropdown` option.
- And then for the tabindex bit, just wrap the line that is already there with the same guard e.g. `if (this.telInput.disabled) [set tabindex to 0]`

Thanks

loicplaire:
Thanks @jackocnr for your feedback, it all makes sense and should be all good now.

jackocnr:
Thanks Loic! I've manually merged your commits.

PR #1281, opened 2021-10-25
[1280] Update libphonenumber in order to accept new formats like 078*…

alexandrupuscu:
…***** and 079****** on RO country


jackocnr:
I've just released a new version (v17.0.14) which is built against the latest version of libphonenumber. Thanks.

Issue #1266, opened 2021-09-09
Attribute "combox" remain when setting allowDropdown to false:


Crocsx:
The attribute `role="combox"` should be set on element when 

> A composite widget containing a single-line textbox and another element such as listbox or grid, that can dynamically pop up to help the user set the value of the textbox.

When specifying `allowDropdown: false`, the attribute `combox` remain on the input, even tho there is no drop down or popping element on the input. This cause Lighthouse to complain, and minor accessibility problems.

![2021-09-09_11-26](https://user-images.githubusercontent.com/6420140/132612353-9ead8d0f-56bb-454d-94f1-34638eb5f954.png)

Exemple => 

https://stackblitz.com/edit/angular-international-telephone-input-zkrosn?file=src/app/app.component.ts

 `role="combox"` should disappear when `allowDropdown` is set to `false`

PR #1284, opened 2021-10-29
#1266 Remove innapropriate attribute when dropdown is disabled.

Crocsx:
I believe there is no way to updated the setting once initialized, so this should be the only change required


jackocnr:
Thanks for this. I have just merged your commits manually.

PR #1317, opened 2022-05-12
1315 - Add visually hidden <label> to combobox control.

### LINKED ISSUE #1315: [a11y] - provide an accessible name to interactive combobox element
When a user accesses the country combobox directly, there is no indication that they're interacting with a phone number input.  They're simply told that they can select a country.  In order to provide enough context so that the country selection element makes sense on its own, we should add an `aria-label` to the interactive element.

### Expected behaviour
A/T announces:

> Telephone country code combo box <the selected country, plus [country code] >

### Actual behaviour
A/T announces:

> <the selected country, plus [country code] combo box>

The difference is that A/T will announce that the selected country is for a "Telephone country code".  Please feel free to bikeshed the exact verbiage to be used, but we should definitely add the additional context.

Please let me know what you think, if this plan sounds good, then I'm more than happy to open a pull request.  Thanks so much for your consideration!

LukeLeber:
This changeset includes the following:

- Adds an `id` attribute to the combobox control
- Adds a `<label>` element that acts as the accessible name with the text `Telephone country code`
- Adds a style that visually hides the `<label>` element to preserve B/C

My only other concern is that the label text is not internationalized properly, but there are other instances of this as well. 
 Bringing full i18n support to the project could likely happen in a follow-up. :slightly_smiling_face: 

Thanks for your consideration!

Closes #1315


mryand:
@jackocnr Can you provide any input on the status of this PR?

jackocnr:
https://github.com/jackocnr/intl-tel-input/issues/1315#issuecomment-1497863987

PR #1323, opened 2022-06-03
updating alreadySelected logic to not update the flag when in separateDialCode mode and country already selected

jonareno:
I believe this is a bug and I'd like to open the discussion with a pull request. 
On the latest version of intl-tel-input, if you change the country from US to another country with the same code (Canada, America Samoa, etc), and start entering a number, the country picker switches back to US. This doesn't feel like a desired behavior. 
In my case, I am using the picker in separateDialCode mode, and nationalMode, and I do not expect users to be entering their country dial code. I expect the dial code to come from the picker. I understand my use case might be unique. Feedback is welcome!

### DIFF
--- a/src/js/intlTelInput.js
+++ b/src/js/intlTelInput.js
@@ -134,8 +134,8 @@ class Iti {
       this.promise = Promise.all([autoCountryPromise, utilsScriptPromise]);
     } else {
       // prevent errors when Promise doesn't exist
-      this.resolveAutoCountryPromise = this.rejectAutoCountryPromise = () => {};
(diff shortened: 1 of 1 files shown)


jackocnr:
Apologies for the delay, but life has been getting in the way! But I have some time now to look into this.

So I just tested it and I can't reproduce your issue. If I select Canada, and type a valid Canadian area code e.g. "506" then Canada stays selected. I also tried with separateDialCode enabled and got the same result. Are you sure you're typing a valid number? If you still think this is an issue, please can you provide a valid non-US number that you can enter and see it switch back to US.

jackocnr:
Closing due to inactivity. But please do let me know if you're still experiencing this issue.

PR #1360, opened 2022-09-27
Fix some a11y issues

anthonydillon:
## Done
- Added `lang` attribute to the `html` throughout the project
- Updated the demo button styling to pass AA contrast checking
- Added `aria-label` to the example mark-up

## QA
- Run `npm install`
- Open the demo.html
- Using the [WAVES browser plugin](https://wave.webaim.org/extension/) check there are no issues
- Suggest reviewing the code with `?w=1` in the URL to hide whitespaces changes which are a result of formatting the codebase.


anthonydillon:
Closing due to inactivity.

PR #1389, opened 2023-01-12
Remove unicode from regex

timbreitkreutz:


timbreitkreutz:
Ugh so sorry.  Wrong fork.

PR #1397, opened 2023-02-15
separateDialCode should (a) set autoHideDialCode true and (b) be independent of nationalMode

johnnyshields:
This PR makes two changes on the same line re: `separateDialCode` option.

a) If `separateDialCode` is true, it should **not** insert the dial code into the input, i.e. `autoHideDialCode` should be **true**. The pre-existing code comment and the actual logic do not agree, so I've corrected it--I believe the pre-existing logic is a bug.

b) `separateDialCode` should not influence `nationalMode`, and users should be able to set it as per their preference. The default should be that nationalMode is enabled.

The reason is, even if the dial code is shown, users still *think* in terms of national format. As an example, in Japan national has a leading zero, and users expect to enter phone this way even when `+81` is shown:
image

If users want a "purist" E.164, they can set `nationalMode` to false. (In the pre-existing logic, there was no way to set it to true!)

(I am using this code in production and users are happy.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ Prevent the user from entering invalid characters. Unfortunately this had to be 
 
 **separateDialCode**  
 Type: `Boolean` Default: `false`  
-Display the country dial code next to the selected flag so it's not part of the typed number. Note that this will disable `nationalMode` because technically we are dealing with international numbers, but with the dial code separated.
+Display the country dial code next to the selected flag so it's not part of the typed number.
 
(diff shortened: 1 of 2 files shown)


jackocnr:
Thanks for the input. This actually led me to reconsider the autoHideDialCode option, which was a bit confusing, as you pointed out, and in the end, I decided to rename it autoInsertDialCode which makes the code a lot clearer I think, so now that option clearly controls both the auto insertion and the auto removal of dial codes.

And I agree about separating nationalMode and separateDialCode, tho it required a bit more work in the end, which I have done separately. I've just pushed this to master. Let me know your thoughts. Thanks again.

PR #1407, opened 2023-03-16
Update intlTelInput.scss to fix browser copiability issue (Guidewire Software company)

ElenaShevchenko:
-moz prefix should be removed because it is outdated and it is course of browser capability issue https://caniuse.com/?search=-moz-box-sizing


jackocnr:
Thanks

Issue #1419, opened 2023-04-25
A11y issue with selecting country option on Chrome


grzegorz-romanczuk:
A11y bug

### Steps to reproduce
1.  Go to https://intl-tel-input.com/ on Chrome
2.  Use VoiceOver and navigate to the country dropdown
3.  navigate between options using keyboard

### Expected behaviour
VoiceOver should read currently focused country. 

### Actual behaviour
VoiceOver skips reading currently focused country and reads the selected one.

### Initialisation options
Library demo

![image](https://user-images.githubusercontent.com/97388222/234282863-b56247fa-3bf1-45d8-ae1b-7f6064507a5a.png)

jackocnr:
Hmm I'm not familiar with VoiceOver. When navigating with the keyboard, how do we communicate to VO which is the highlighted country?

jackocnr:
Yes well the W3 example that we use as the basis for our aria tags actually doesn't include aria-owns so I would be happy to remove it. I would be open to a pull request that makes this change. Thanks

Issue #1415, opened 2023-04-20
A11y issue with country dropdown 


ElenaShevchenko:
A11y
### Steps to reproduce
1.  Set "separateDialCode" to true, country dial code next to the selected flag is appear
2. Choose Finland in a country list 
2.  Use NWDA screen reader and navigate to the flag dropdown

### Expected behaviour
NWDA should announce something like:  “Finland +358 combo box  collapsed”.

### Actual behaviour
NVDA announces “Finland +358 combo box +358 collapsed”. **SO the code is announced twice** 

### Initialisation options          
            separateDialCode: true,
            autoHideDialCode: true,
            allowDropdown: true,
            nationalMode: false,
            formatOnDisplay: true,
            
This issue is appear becouse country code is duplicate in "title" attribute
![image](https://user-images.githubusercontent.com/11269499/233320275-88c0f117-d0f9-491f-976b-6eff75d816fc.png)

jackocnr:
Hmm what would you propose?

ElenaShevchenko:
I think we can remove "+358" from the title for iti__selected-flag div in case separateDialCode: true,
only "Finland" should stay

PR #1421, opened 2023-05-05
1415  and 1419 screen reader enhancements

danman01:
1419 - fixes screen reader not reading select box option

closes https://github.com/jackocnr/intl-tel-input/issues/1419

1415: remove duplicate country code from screen reader

closes https://github.com/jackocnr/intl-tel-input/issues/1415


jackocnr:
Thanks so much for putting this together @danman01.

My only feedback is that when separateDialCode is disabled, we still want to include the dial code in the title.

danman01:
@jackocnr Updated! Let me know if you think the specs are in the right place? After working on it, it looks like the "Unknown" title scenario won't ever get hit, but there might be some case where we still need it in there.

jackocnr:
This looks really great, thank you! Any chance you could build the JS as well? Then we can merge.

https://github.com/jackocnr/intl-tel-input/blob/master/.github/CONTRIBUTING.md#changes-to-the-plugin

danman01:
@jackocnr I ran the build js command and committed the results

jackocnr:
Thanks again

PR #1430, opened 2023-06-16
Fix separate dial code padding calc on right for RTL

pacotole:
With a separate dial code, in RTL languages, input padding is added to the left and should be added to the right where the flag and country code is.


jackocnr:
Thanks for putting this together. Apologies for the delay, I've had a lot on, but I should be able to get to this in the coming week.

jackocnr:
This is great thank you. Merged manually 👍

PR #1432, opened 2023-06-19
Updating flag images region-flags

t-bonfill:
- upgrade region-flags to 1.2.1 (direct url because latest tag not available on NPM) to upgrade deprecated flags (E.g Martinique's flag recently been modified : https://en.wikipedia.org/wiki/Flag_of_Martinique )


jackocnr:
Thanks for putting this together. Apologies for the delay, I've had a lot on, but I should be able to get to this in the coming week.

t-bonfill:
Great, thanks for your work !

PR #1453, opened 2023-08-09
Introduced option : afterSetNumber

Surdok:
Hello.

I have put up a callback function that could be passed to the options, so that whenever I call the setNumber, the callback function gets called, and does whatever thing I pass to it.

I mainly use this callback to format the received mobile numbers, as extra +'s appear, or some duplicate country code gets set from the setNumber.


jackocnr:
> extra +'s appear, or some duplicate country code gets set from the setNumber

These issues sound like bugs to me that we should fix separately. Can you provide clear steps to reproduce any issues you're facing?

Surdok:
What I actually aim in doing through using the library, is that I want to initialize the country code dropdown, alongside the number input itself.

What I am facing in reality is that the data coming from our backend sometimes contain missing data and un-refined data. Which leads to the library not properly initializing properly.

You can have a look at this short gif showing on some un-intended behaviour.

I am trying to initialize the input field with an incomplete number (complete number being one with 8 digits)
it successfully switches to the correct country code in the drop down, but it also inserts the country code wrongfully into the input field.

Also it shows space between the numbers, which is not needed, and gets stored in the database. Later on when using the number with a space, it shows the number with space, which is not wanted.

https://github.com/jackocnr/intl-tel-input/assets/24224586/f21a8936-b995-41f0-b338-4d1f4717cbe6

jackocnr:
Ok, so if I understand you correctly, you have 2 problems:

First, if you call setNumber with an incomplete number like "+9731220" then it correctly sets the selected country, but then leaves the international dial code in the input, which you don't want. But if we just removed the intl dial code, that would leave the number in an even more confusing state, as the number would look like it was in national format (without the intl dial code), except in several countries they add an extra digit at the beginning in national mode e.g. in the UK they add a 0, so if you had an incomplete number "+447947" and you selected the UK flag and put "7947" in the input, that wouldn't make any sense and would look wrong. I think the best thing to do in this situation is display the closest thing you have to a valid number, which is the full thing including the intl dial code. I don't see any harm in leaving it there - do you? My advice to you would be to run a script to clean up your database - if you have bad numbers, then delete them - they're useless anyway, right?

Second, if you call setNumber with a valid number like "+97332202112" then it correctly sets the selected country, and  removes the international dial code, but it also formats it with a space in the middle, which you don't want. In this case I would just say: that is the standard national format for that country - what's the problem with the space?

I'm afraid that unless I'm missing something, I really don't see the need to add any new features to the plugin just for these use cases. I'll close this for now, but I'm happy to keep discussing if you'd like.

jackocnr:
> Also it shows space between the numbers, which is not needed, and gets stored in the database.

By the way, you shouldn't be storing any spaces to your database. You should get the number using the getNumber method, which would give you the number in standard E164 format.

PR #1463, opened 2023-08-22
feature: implement option "useFullscreenPopup"

### LINKED ISSUE #1326: Is there a way to make the plugin to draw the mobile layout on desktop screens?
Is there a way to make the plugin to draw the mobile layout on desktop screens?
I'm using the plugin in small windows (iframe) and would like to show mobile layout (when dropdown is fullscreen).
In the code I see you're using `navigator.userAgent` check:

> this.isMobile = /Android.+Mobile|webOS|iPhone|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent);

but `navigator.userAgent` is read only and I can't just temporary set it to `iPhone` for example during initialisation.

Any other options how to achieve that?

tw-360vier:
closes #1326


jackocnr:
Additionally, what do you think about adding a 2nd check to the default value, so it's something like this:

`useFullscreenPopup = userAgent mobile test || window.innerWidth < 500`

So by default, we show the fullscreen popup on any mobile device, or on any viewport thinner than 500px. I think this makes sense because even on a desktop device, if you've got your browser window really thin, it looks better having the fullscreen popup. And also we have a breakpoint in the CSS at 500px, below which the dropdown suddenly gets very thin and doesn't look as good, which is not easily fixed (for complicated reasons, which I'd be happy to discuss separately if you're interested) - so for this reason as well, in this case I think it would be better to default to the fullscreen popup.

tw-360vier:
@jackocnr i have updated the pr to fit your remarks.

jackocnr:
I merged your commits manually. Thanks again.

PR #1484, opened 2023-12-01
Update data.js

zoe-elk26:
Please remove western sahara from the Update data.js
the list of GSM numbers shows Western Sahara, the entity belonging to the Polisario front of RASD. This is a very serious mistake and affects the territorial unity of Morocco. Please remove Western Sahara from your GSM list


jackocnr:
Thanks

knyghty:
Hi @jackocnr,

I'm not sure why this was merged. From my limited understanding, Western Sahara is recognised by the UN and only a handful of countries recognise Moroccan control over it. It seems like the person requesting this may have some bias in wanting it to be merged (and I don't see any other GitHub activity from them). Personally I'm not from any of these places and don't have a horse in the race so to speak. But it seems like a bad precedent to me to remove recognised territories because someone doesn't want it to be there. Can you share your thinking on this?

jackocnr:
Thanks for looking into this @knyghty. When the issue was raised, I did a quick search and saw on the [Western Sahara wiki page](https://en.wikipedia.org/wiki/Western_Sahara) that in 2020 the USA recognised Moroccan sovereignty. This coupled with the fact that Google's phone input (e.g. when you use Google Contacts) no longer includes Western Sahara, so I thought ok the situation has changed, let's remove it.

But since you challenge it, I have read more, and you're right that Western Sahara is recognised by the UN as a "non-decolonized territory" and [this wiki page](https://en.wikipedia.org/wiki/Political_status_of_Western_Sahara) says "Under international law, Western Sahara is not a legal part of Morocco". Also I found that Stripe's phone input do still include Western Sahara. So I've re-added it for now.

I welcome further discussion here, but from now on, when making any claims, please link to an authoritative source.

YazidKHALDI:
Request to Remove Western Sahara as a Separate Entry from the intl-tel-input Plugin

The current configuration of the intl-tel-input plugin lists Western Sahara separately with the +212 phone code. However, +212 is exclusively assigned to Morocco and is used by Moroccan telecom companies. It is important to note that Western Sahara is recognized internationally, including by the United States, France, and Spain, as part of Morocco. Historically, Spain colonized this territory after the Berlin Conference of 1884, but it remains an integral part of the Kingdom of Morocco.

Given this historical context and current international recognition, it would be more accurate to remove Western Sahara as a distinct entity in the plugin, aligning with the reality that Morocco and Western Sahara share the same telephone code under Moroccan jurisdiction.

Thank you for considering this update to improve accuracy and avoid geopolitical confusion.

knyghty:
@YazidKHALDI so, I understand the Moroccan nationalist point of view - this is clear. My question is more about the UX. If you consider yourself to live in Western Sahara and you try to add your phone number, it's quite reasonable to try to find your country in the list, no?

YazidKHALDI:
@knyghty  
Thank you for your reply. I understand the concern from a UX perspective. However, it's important to note that anyone living in Western Sahara uses Moroccan GSM companies with the +212 phone code, holds a Moroccan ID, and interacts with Moroccan authorities for all official matters. For example, tourists visiting cities like Dakhla or Laayoune arrive at Moroccan airports and are processed by Moroccan authorities. This reflects the reality on the ground, where Western Sahara is fully integrated into Moroccan infrastructure.

As for those who consider themselves "Western Saharan" in a different sense, most are living in refugee camps in Algeria and are supported by the Algerian government for geopolitical reasons. Algeria is, in fact, the only country actively supporting the creation of a separate state for these individuals.

In this context, listing Western Sahara separately in the country dropdown can cause unnecessary confusion. All phone numbers in this region use Morocco’s +212 code, and users residing there are accustomed to identifying with Morocco in all aspects of life, including telecommunications.

I hope this clarifies why it would improve the user experience to remove Western Sahara from the list.

knyghty:
Everything you just said on both of your comments sounds like nothing but propaganda to me. You talk about Morocco and about Algeria but not the people actually living there. I think we are missing their perspective here, don't you think?

YazidKHALDI:
I believe we are approaching this from very different perspectives. My main point was to highlight the practical user experience and the existing infrastructure in the region, which aligns with Morocco’s telecommunications and governance.

If every region where there is a separatist movement or external influence were listed separately, we would need to add hundreds of contested areas globally. We would see entries like Catalonia, Crimea, Quebec, South Ossetia, Donetsk, Luhansk, and many more.

With that said, I respect your viewpoint, and I’ll leave it at that. Thanks again for your time and consideration.

jackocnr:
> listing Western Sahara separately in the country dropdown can cause unnecessary confusion. All phone numbers in this region use Morocco’s +212 code, and users residing there are accustomed to identifying with Morocco in all aspects of life, including telecommunications.

Can you explain why it would be confusing? If I'm from Western Sahara and I open the country dropdown, I can either type "Morocco" or "Western Sahara", either way, I will find what I'm looking for. I think they would be able to enter their phone number under either country without issue. So I don't understand what is the downside to listing Western Sahara?

We do the same thing for islands off the UK: if you open the country dropdown and type +44 you will see the UK along with 3 other territories that use that dial code. Most people would consider those as part of the UK, but they have their own identity, so they get their own territory, even though they use the same dial code as the UK.

For me, it basically comes down to: is there a chance a user from that territory might search for it? If so, it's better to give them what they're looking for. (provided it meets our [criteria](https://github.com/jackocnr/intl-tel-input/blob/master/src/js/intl-tel-input/data.ts#L3-L7) for listing a territory).

PR #1485, opened 2023-12-06
Fixing non compliant divider in country dropdown

stuart-johnson-calendly:
Divider li in the country dropdown, between "preferred" countries and the rest of the list, uses the: role='separator' attribute/value. This does not comply with the aria a11y standards, as separators are not allowed in listboxes. I've removed the role, and made the divider aria-hidden='true', to remove it from the a11y tree, since there isn't a particularly good way to implement it that is compliant with a11y standards.


stuart-johnson-calendly:
@jackocnr So, I'm having trouble finding a specific statement about it NOT being allowed, but the aria docs/specification seem to infer that it is expected to only be between menuitems specifically, and therefor not allowed between anything else?

This thread here seems to be confident, and it's from the axe-core project:
https://github.com/dequelabs/axe-core/issues/3938

jackocnr:
Thanks. I merged your commit manually 👍

stuart-johnson-calendly:
@jackocnr Awesome, thanks! Any chance for a quick patch release with this? I'm in the middle of a big a11y compliance project and this is one of my outstanding issues.

jackocnr:
Done

PR #1488, opened 2023-12-19
fix: updating to verify the existance of window or navigator to work on ssr

paulorievrs:
This will fix an error when using the library with NextJS or other SSR frameworks, where, it wasn't verifying the window and the navigator before using. See the print below the check the error that WAS happening before this fix.

image


jackocnr:
Thanks for this

jackocnr:
Released as v18.3.4

Issue #1481, opened 2023-11-27
Shadow DOM and Mobile View issue


skuzmin:
### Steps to reproduce
1. Create basic [WebComponent](https://developer.mozilla.org/en-US/docs/Web/API/Web_components) with intl-tel-input plugin
2. Switch to mobile view and refresh page (enable mobile mode)
3. Click open list of countries

### Expected behaviour
Pop-up(modal/dialog) with list of countries should be visible

### Actual behaviour
There is no Pop-up(modal/dialog). Because in mobile mode we have "iti-mobile" class on document.body(hardcoded)
<img width="353" alt="image" src="https://github.com/jackocnr/intl-tel-input/assets/8438624/48d30476-ce59-46c1-9f5c-01d210714f44">
And this doesn't have effect on incapsulated content inside shadowDom.

The same story with container, but in this case I can specify container inside shadowDOM. It would be great to have option to select element where we can add class for mobile view/mode in the same way as we have for container. If I add class manually to shadow dow element then everything works.

### Initialisation options
utilsScript: https://cdn.jsdelivr.net/npm/intl-tel-input@latest/build/js/utils.js
dropdownContainer: element inside shadow DOM

jackocnr:
I see we reference `document.body` 5 times in the [source JS](https://github.com/jackocnr/intl-tel-input/blob/master/src/js/intlTelInput.js) so we would need to replace all of these. I propose a new bodyContainer option, which defaults to `document.body`. Would that solve your issue?

jackocnr:
Well we're talking about the same thing - adding a new option to pass in a container element. You want this container element to take the iti-mobile class, but while we're doing that, we may as well use it for the other instances where we use document.body in the code, to keep it encapsulated within the shadow DOM. e.g. when you open the dropdown, the code currently uses document.body.scrollLeft to figure out where to position the dropdown, and in your case, this will need to be relative to the shadow DOM instead of the whole document.

I don't have time to implement this myself right now, but I would be open to a pull request.

oprichnik:
Behaviour can be reproduced by changing user-agent in your browser.

If you don't need the popup, you can disable it with the `useFullscreenPopup` option.

mimoudix:
did you fix this issue ? same btw

skuzmin:
@mimoudix sorry ,I haven't had time to take a look into it, because of holidays :)

but I added temporary fix to my app (web component) to be able to use the lib in mobile view.
[...]

PR #1497, opened 2024-01-04
Add shadowDomContainer option, update read.me

skuzmin:
Related issue: #1481 

Also add read.me description of this property.
It should be backward compatible.


skuzmin:
Fixed in v18.4.0

Issue #1509, opened 2024-01-18
Broken Search Broken module Broken all


mikadev:
By playing around with multiple instance i get first this one https://github.com/jackocnr/intl-tel-input/issues/1505 the the second search is completely broken (see capture)

![image](https://github.com/jackocnr/intl-tel-input/assets/3314041/7fe192f2-3762-4f96-b1bb-c706621c00f3)

this is also reproductible here => https://intl-tel-input.com/examples/multiple-instances.html (all browsers **and only with multiple instances**)

![image](https://github.com/jackocnr/intl-tel-input/assets/3314041/a251b3db-4d0e-4e3a-b157-d724ce3a91d3)

this bug is critic because the module become frozen 

Broken code => https://github.com/jackocnr/intl-tel-input/blob/50b00b2d0195e80b0c39c2d878a7400709ae9cbb/src/js/intlTelInput.js#L1360

mikadev:
this is a nightmare plugin anyone know a valid alternative please ?

lekoala:
no worries this is easily fixed

mikadev:
A ok thank you :)

jackocnr:
Apologies for the issues. I've just made a series of big changes to the plugin all at once and it was inevitable there would be a few teething issues. We'll get them ironed out shortly.

mikadev:
Hello @jackocnr thanks for you supporting this plugin (is still broken this morning (19.2.5)

1. Search country in first input
2. Search country in second input
3. Then open the search in the first one again

![image](https://github.com/jackocnr/intl-tel-input/assets/3314041/342d6875-a19f-47f6-baa4-3b593e08e3e7)

jackocnr:
Fix released in v19.2.6

mikadev:
Thanks !

PR #1511, opened 2024-01-18
check nextItem

lekoala:
Fixes https://github.com/jackocnr/intl-tel-input/issues/1509


jackocnr:
Thanks for this, but in the end I chose to fix it in a slightly different way: https://github.com/jackocnr/intl-tel-input/commit/374d510fb8692fed930cd99fc9db11103fa8c296

PR #1524, opened 2024-02-13
Fix/dropdown position when using dropdown container option

Fectorez:
# Type
Fix

# Context
When using `dropdownContainer` option, the dropdown should be below (or above, depending on the input position on the screen) the input.
The dropdown X (left) and Y (top) are computed using the current viewport, and are applied using `top` and `left` CSS properties.

## Example

Given:
- Input X (left): 50px;
- Input Y (top): 100px;
- Input height: 30px;
- And no scroll

Then: the dropdown should have:
- X (left): 50px;
- Y (top): 130px.

# Problem
Setting element position using `top` and `left` CSS properties is **not relative to the viewport**, but relative to the **nearest ancestor having a position other than `static`**.
Consequently, if the `dropdownContainer` or one of its ancestors has a position other than `static`, then the dropdown will not be positioned correctly.

Example with the dropdown container having `margin: 50px`:
image

# Steps to reproduce

Run this:
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Document</title>
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/intl-tel-input@19.2.16/build/css/intlTelInput.css">
    <script src="https://cdn.jsdelivr.net/npm/intl-tel-input@19.2.16/build/js/intlTelInput.min.js"></script>
</head>
<body>
                <!--    Important style     -->
    <div style="position:relative; margin: 50px" id="container">
        <input type="tel" id="phone">
    </div>
    
    <script>
    const input = document.querySelector("#phone");
    window.intlTelInput(input, {
        utilsScript: "https://cdn.jsdelivr.net/npm/intl-tel-input@19.2.16/build/js/utils.js",
        dropdownContainer: document.getElementById('container')
    });
    </script>
</body>
</html>
```

# Fix
First, we set `position: relative` to the `dropdownContainer`. That styling should be written in the `dropdownContainer`  part of the documentation, because that alters a host page's element.
Then, we set:
- `left: 0` ;
- `top: inputHeight` if the dropdown is below the input ;
- `bottom: dropdownHeight + inputHeight` if the dropdown is above the input.

(Only the first/older commit should be reviewed; the second/newer is a build).

# Other possible solution
Put the dropdown as a direct child of `body`: then the initial `top` and `left` would be correct. However, we will have to add accessibility to correctly link the input and the dropdown.

# Notice
Please do manual tests before merging, because I don't know very well this project, and I may have cause a regression. And be free to do changes or create another PR if you have a better solution.


jackocnr:
Thanks for reporting this, and giving it so much thought. The issue with your solution is that it wouldn't work if (A) there were other elements inside the container before the tel input, or (B) the container was not an ancestor of the tel input at all e.g. it was just a div at the end of body etc. But your input led me to a much simpler solution - just use `position:fixed` instead. This fix is included in the new v19.2.17 - give it a try and let me know what you think.

sushantgwr87:
Thanks for fix but calculations seems bit off for me. Top position is setting way far away from div container ID i have set as dropdownContainer.

Here I set - dropdownContainer: document.querySelector('#form-num-input')

I am using Next.js 14 and npm package of this

```
<div
  id="form-num-input"
  class="!relative !block h-fit"
>
  <div class="!relative !block w-full">
    <div class="!relative !block w-full">
      <input
        class="hidden"
      />
      <div
        class="iti iti--allow-dropdown iti--show-selected-dial-code iti--show-flags iti--inline-dropdown"
      >
        <div class="iti__flag-container">
          <div
            class="iti__selected-flag"
            role="combobox"
            aria-haspopup="listbox"
            aria-controls="iti-2__country-listbox"
            aria-expanded="true"
            aria-label="Selected country"
            tabindex="0"
            title="India"
            aria-activedescendant="iti-2__item-bn"
          >
            <div class="iti__flag iti__in"></div>
            <div class="iti__selected-dial-code">+91</div>
            <div class="iti__arrow iti__arrow--up"></div>
          </div>
        </div>
        <input
          compname="PhoneNumber"
          phoneformat="1"
        />
      </div>
    </div>
  </div>
  <div
    class="iti iti--container iti--inline-dropdown iti--country-search"
    style="top: 606.258px; left: 70.7734px"
  >
    <div class="iti__dropdown-content" style="width: 375px">
      <input type="text" class="iti__search-input" placeholder="Search" />
      <ul
        class="iti__country-list"
        id="iti-2__country-listbox"
        role="listbox"
        aria-label="List of countries"
      ></ul>
    </div>
  </div>
</div>
```

jackocnr:
@sushantgwr87 if you can find the minimal set of code that reproduces the problem and put that in a CodePen, then I'd be happy to take a look. Here is [an example pen](https://codepen.io/jackocnr/pen/DyrVbw) with the plugin up and running - feel free to fork this if it helps.

Fectorez:
@jackocnr Thank you very much for your fix!

jackocnr:
@sushantgwr87 btw I notice you're using React - you may be interested in our new [React component](https://intl-tel-input.com/examples/react-component.html).

dejan-lukic:
I'm also seeing issues with the `top` calculation, the dropdown ends up quite far from the input. Cannot produce a minimal reproduction at the moment, but will do so if I get the chance

jackocnr:
@dejan-lukic this issue is about the dropdownContainer option - are you using that option, as I didn't see it in the codepen? If not, please can you create a new issue.

Also, strangely, when I view your codepen, I cannot reproduce your issue. I load the page, scroll down a bit, and then click the dropdown, and it appears in the correct position. When you create a new issue, please state which browser/version you're using. And also: when you say you've edited the original codepen - can you please state exactly which changes you have made, as I can't really see any, apart from setting height on the body. Can you reproduce the issue in [the vanilla codepen](https://codepen.io/jackocnr/pen/DyrVbw)? And also please update it to use the latest version of the plugin. Thanks.

dejan-lukic:
@jackocnr sorry, I don't know what happened, seems the JS in the codepen didn't save... I've just tried to save it again, hopefully this time it works. The only changes I made was updating the body height and setting the `dropdownContainer` option to `document.body`.

For reference, this was working as intended in `v17.0.18` (version we were using before bumping up to `v21.1.4`). 

I've updated the codepen and bumped the plugin version to `21.1.4`, let me know if that's now working

jackocnr:
Thanks I can repro now. I'll look into this when I get a second.

jackocnr:
@dejan-lukic thanks for raising this. Surprised it hasn't come up before!

I've just released a fix in v21.2.4 - let me know if that works for you.

dejan-lukic:
@jackocnr thank you very much! I’ll test out the new release in the morning and post back here

PR #1529, opened 2024-02-20
fix: Handle brackets in hidden input name when adding '_country' suffix

carlssonemil:
The adjustment I'm proposing addresses a compatibility issue with Ruby on Rails when input names include brackets, like `user[phone]`. Currently, appending `_country` to such input names results in `user[phone]_country`, which typically doesn't pose a problem. However, in Rails, this syntax disrupts the conventional structure Rails expects for form inputs and parameters, leading to potential parsing errors or unexpected behavior. Rails relies on specific naming conventions to organize and process form data, particularly for nested attributes, to ensure the parameters' proper handling and mapping to model attributes. This structure is crucial for maintaining the framework's RESTful design principles, which include predictable routing, parameter organization, and URL generation[^1][^2].

To resolve this, the modification I suggest more closely aligns with Rails' expectations by inserting the `_country` suffix within the brackets for nested attributes, changing `user[phone]_country` to `user[phone_country]`. This approach preserves the conventional parameter structure expected by Rails, avoiding potential issues with parsing and handling form data. Additionally, I've developed tests to verify that this solution effectively addresses the identified issue, ensuring compatibility with Rails' form and parameter handling conventions.

[^1]: [Action View Form Helpers](https://guides.rubyonrails.org/form_helpers.html)
[^2]: [Using Rails for API-only Applications](https://guides.rubyonrails.org/api_app.html)


jackocnr:
Thanks for putting this together, but after reflecting on this for a while, I think I'd rather avoid making any assumptions about what people want. e.g. what about situations where there are multiple square parentheses.

Instead, what I'd rather do, for now, is to allow people to return an object from the `hiddenInput` function, with 2 keys: `phone` and `country`, with the values being set to the relevant names e.g. 

```js
hiddenInput: function() {
  return {
    phone: "my_phone_input_name",
    country: "my_country_input_name",
  };
}
```

And so all we'd have to do in the plugin code is check if they return an object or a string e.g.

```js
const result = hiddenInput(telInputName);
const isObject = typeof result === "object";
const phoneName = isObject ? result.phone : result;
if (isObject && result.country) {
  // insert country input
}
```

How does that sound?

jackocnr:
@carlssonemil it would be great if you could put this together and submit a new pull request, thanks so much!

PR #1542, opened 2024-03-09
limit the number of files to be included in the build

### LINKED ISSUE #1541: Decrease npm package size
Users of the `intl-tel-input` npm package need the CSS and JavaScript utilities located in the `build` directory, and do not need to install any of the dependencies, submodules and other tooling of the repository.

Currently, the npm package copies the content of the whole github repository at each install, including the content of `libphonenumber` dependency, which consumes more than 100 MB of disk space.

This is problematic in multiple ways:

* the package takes a long time to download whenever installing with a slow Internet connection
* the package takes a long time to decompress on low spec, slow hardware
* the size of the package might create challenges when using in programs installed on low spec hardware
* consequently, the programs using the package suffer from the same problem: they are slower to install and deploy, and consume a lot of disk space unnecessarily.
* all the overhead has a negative impact on the environment as it causes unneeded energy consumption - even if this impact is negligible compared to other factors.

### Steps to reproduce

1.  Navigate to an empty directory
2.  Execute the command `npm install intl-tel-input`
3.  Check the installed packages and the space they use `du -sh node_modules/*`
4. Check the number of files installed `find node_modules/ -type f | wc -l`

### Expected behaviour

* The only node.js module installed is `intl-tel-input`
* The number of files installed is small, definitely less than a hundred files
* The disk space used by the module is small, ideally less than a megabyte

### Actual behaviour

* 8 packages are installed in addition to `intl-tel-input`
* 3975 files are present inside `node_modules` directory
* `node_modules` directory consumes 122MB of disk space

Current output of `du -sh node_modules/*`:

```
117M	node_modules/intl-tel-input
32K	node_modules/js-tokens
36K	node_modules/loose-envify
20K	node_modules/object-assign
128K	node_modules/prop-types
380K	node_modules/react
4.4M	node_modules/react-dom
56K	node_modules/react-is
156K	node_modules/scheduler
```

### Initialisation options

N/A

### Proposed solution

1. Move `react`, `react-dom` and `prop-types` from `dependencies` to `devDependencies` inside `package.json`.
2. Use the `files` property inside `package.json` in order to specify which files should be included in the npm package.

For example, the following setup decreases the size to 25 files, using 1.2MB (100x reduction):

```
  "files": [
	  "build/*",
	  "CHANGELOG.md",
	  "LICENSE",
	  "package.json",
	  "package-lock.json",
	  "README.md",
	  "screenshots/*"
  ],
```

mef:
This is a proposal to fix #1541.

Changes:

* Move all dependencies to `devDependencies` instead of `dependencies`, so that they are not included when running `npm install intl-tel-input`.
* Only include the build and the documentation inside the npm package.

This dramatically decreases the package size, making it faster to install, and consuming 100x less disk space (1.2MB instead of 122 MB).

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,8 +61,20 @@
     "prettier": "^2.8.7",
     "region-flags": "https://github.com/fonttools/region-flags/archive/refs/tags/1.2.1.tar.gz",
     "sass": "^1.62.0",
-    "time-grunt": "^1.4.0"
+    "time-grunt": "^1.4.0",
+    "prop-types": "^15.8.1",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
   },
+  "files": [
+	  "build/*",
+	  "CHANGELOG.md",
+	  "LICENSE",
+	  "package.json",
+	  "package-lock.json",
+	  "README.md",
+	  "screenshots/*"
+  ],
   "repository": {
     "type": "git",
     "url": "https://github.com/jackocnr/intl-tel-input.git"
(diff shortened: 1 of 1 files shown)


jackocnr:
Thanks very much for this. I've merged the commit manually and added a further commit with some fixes. I'm just waiting for Travis to fix my testing account to check all ok there, and then I will do a new release including this. Thanks again.

PR #1552, opened 2024-03-29
Adds CSS variables suport

anthony0030:
After updating to V20 my code broke.
There is no longer a way to override SCSS variables because the src files were missing from the node_module.
If the variables and calculations are done in CSS this is no longer a problem.

This update:
**1.** Adds CSS variable support
**2.** Updates libphonenumber
~~**3.** Updates images~~
**4.**  Adds Node version files


jackocnr:
Thank you so much for putting this together! CSS variables are definitely the way to go in 2024. I just have a few questions before merging:

- Now that the Sass variables are only being used to populate the CSS variables, they feel a bit redundant - what would you think about getting rid of them altogether? (esp since we no longer include the sass files in the bundle, so there's no way for people to override them anyway). If you agree, I think it's fine to drop the `$prefix` variable (but keep the `iti-` in the variable names), as although it is a nice option, in my opinion, it is not needed and makes the code harder to parse mentally.
- Is there any advantage to having the variables declared in the `:root` scope, as opposed to confining them to the plugin `.iti` scope?
- Finally, with the image update commit - was there a particular reason for this? Have any flags changed?

Thanks again, I look forward to getting this merged.

anthony0030:
Hi @jackocnr,

The reason I kept the SCSS variables was to not need to make any breaking changes.
Some people might be downloading the source to a local lib folder and building it like that.
I had originally removed them locally but put them back so as not to need a major version bump.
_**I am in favor of removing the SCSS variables.**_

The reason for making the prefix changeable is I have seen that practice done in repos like **bootstrap**
_**I am in favor of removing the prefix customization**_

The advantage of declaring the CSS variables in the root scope is it doesn't matter what order the main CSS is loaded.

This order is what I am referring to.
```HTML
  <link rel="stylesheet" href="/styles.css">
 <link rel="stylesheet" href="/intlTelInput.css">
```

If you load up the code in this order and they are scoped to .iti, overeating the variables is harder.
I was thinking of scoping them to root and having the customizations done via the .iti class.

The pros of having the variables in the root It makes the code more foolproof to the order of the CSS being loaded.
This is how I have seen other packages handling the situation.
_**I am not in favor of moving them to the .iti class**_, but will if asked by you.


Regarding the flag update, I misunderstood the documentation as a list of things for a contributor to do to get a merge request. Almost like a way to ensure the external assets stay up to date. The only difference I saw was a slightly smaller file size.
_**If you want I can remove the changes regarding the flags if you want**_

Thanks for making such a useful package

jackocnr:
> I had originally removed them locally but put them back so as not to need a major version bump.

Thanks for the thought. I actually have some other breaking changes lined up, so let's go ahead and remove the SCSS vars and I'll include this change as part of an upcoming v21 release.

> The reason for making the prefix changeable is I have seen that practice done in repos like bootstrap

Yes I spotted that, and again I appreciate the thought, but in this case, I think it's overkill. Let's remove.

> The advantage of declaring the CSS variables in the root scope is it doesn't matter what order the main CSS is loaded.

Sounds good, let's keep it in the root.

> Regarding the flag update, [...] The only difference I saw was a slightly smaller file size

I'd say let's avoid breaking everyone's caches until we really need to, so let's drop this change.

> Thanks for making such a useful package

💐 

Thank you for such a thoughtful and useful PR! 👏

anthony0030:
Hi @jackocnr, Sorry for the delay I had some fires to deal with this morning.
I made the changes that we discussed earlier.

I have a question that relates to the build/docs.

In the contributing documentation, it says:

```
In order to build the project, you will first need to install [npm](https://www.npmjs.org/), and then run npm install to install the project's dependencies (for now you must be using node v16 due to our node-sass dependency). At this point, the included demo.html should be working, if you open it in your browser. Then you should make your changes in the src directory, and be sure to run the build script before committing your changes - see below for more information on this.
```

Locally I tested with the latest stable version of node (v21.7.1) and sass (DART 1.72.0), It worked well.

The sass package now refers to dart-sass and not node-sass. Would you be ok if I made a pull request updating dependencies and migrating to the new SCSS build system and the relevant documentation?

I understand if that is something that you would personally want to do.

jackocnr:
That's now merged into master and will be included in the upcoming v21 release (hopefully later today). Thanks again!

> Would you be ok if I made a pull request updating dependencies and migrating to the new SCSS build system and the relevant documentation?

That sounds great, thanks.

jackocnr:
v21 released 👍

PR #1554, opened 2024-04-03
:bug: Fixes exports

ekwoka:
The exports field only exposed the default file, and not any of the other files many projects depend on like css and utils

This expands it with pattern matched export to allow all files.

This just mimics the old behavior, so I just went with a catch-all


jackocnr:
Thanks

Issue #1558, opened 2024-04-05
Next.js: cannot find module intl-tel-input/react or its corresponding type declarations


qinfengwu90:
### Steps to reproduce
1.  used `yarn add intl-tel-input` to install the latest version 21.0.5
2.  tried both Webstorm and VS Code
3.  also installed `"@types/intl-tel-input": "^18.1.4"` and imported `"intl-tel-input/build/js/types.d.ts"`;
`

### Expected behaviour
It should import the `IntlTelInput` React component

### Actual behaviour
IDE says they couldn't find the module 

<img width="1025" alt="Screenshot 2024-04-05 at 09 40 21" src="https://github.com/jackocnr/intl-tel-input/assets/45974309/5e477567-48e1-4b1e-86b0-b17dc0b13dbd">
<img width="935" alt="Screenshot 2024-04-05 at 09 39 35" src="https://github.com/jackocnr/intl-tel-input/assets/45974309/e44aebf2-70d5-4fef-b733-ef43f03a2712">

Thanks for looking into it

jackocnr:
Thanks for reporting this. Unfortunately, I cannot reproduce the issue - I just created a new Next.js project with `npx create-next-app@latest` (which uses Next.js v14.1.4) and then did `npm install --save intl-tel-input` and then imported the plugin with `import IntlTelInput from "intl-tel-input/react"` and it works fine for me. I do get a linting error saying "Could not find a declaration file for module 'intl-tel-input/react'" (this is a known issue) but the dev build works fine and successfully displays the plugin on the page.

Some questions for you:

- What version of Next.js are you using? (and Node version too)
- Can you share your Next config
- And your TSconfig as well
- And any custom Webpack config too if you have one

jackocnr:
In v21.0.6 I've fixed the "Could not find a declaration file" error.

qinfengwu90:
> Thanks for reporting this. Unfortunately, I cannot reproduce the issue - I just created a new Next.js project with `npx create-next-app@latest` (which uses Next.js v14.1.4) and then did `npm install --save intl-tel-input` and then imported the plugin with `import IntlTelInput from "intl-tel-input/react"` and it works fine for me. ~I do get a linting error saying "Could not find a declaration file for module 'intl-tel-input/react'" (this is a known issue) but the dev build works fine and successfully displays the plugin on the page.~ (Fixed in v21.0.6)
> 
> Some questions for you:
> 
[...]

PR #1561, opened 2024-04-07
fix: Add missing type definition. Allows to autoconfigure typescript compiler

mpalourdio:
Hi,

The `type` entry has been removed from `package.json` in v21.0.6. This should be re-introduced for consumers. Without it, we need to modify tsconfig.json at userland. Which has already been broken by renaming the file between 21.0.5 and 21.0.6

```json
"paths": {
      "intl-tel-input": [
        "node_modules/intl-tel-input/build/js/intlTelInput"
      ]
    },
```
This is needed today, but shouldn't (I do not use React, but angular FWIW)

See one of my comment here :  https://github.com/mpalourdio/intl-tel-input-ng/pull/54#discussion_r1554837350

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,5 +91,6 @@
     "./data": "./build/js/data.js",
     "./react": "./react/build/IntlTelInput.js",
     "./*": "./*"
-  }
+  },
+  "types": "./build/js/intlTelInput.d.ts"
 }


jackocnr:
I removed this because I read that if the types file lives next to the JS file and has the same name, then that should be enough for the code editor to automatically pick it up, and you shouldn't need a types entry in package.json. Additionally, we now have multiple types files, and it seemed wrong to specify only one of them here. With that said, I'm willing to try re-adding this if it solves a problem for you, and we can review if it causes any issues for other people (e.g. who are importing the react component, which has its own types file).

jackocnr:
Released in v21.0.8

mpalourdio:
Thanks, just tried 21.0.8 and does the job for me et least. I think you can handle it in other ways of you want to have multiple types definition, like publishing them in the `@types` dedicated repository, but that's indeed more work.

I suppose 21.0.8 will fix https://github.com/jackocnr/intl-tel-input/issues/1558 too

Thanks.

mpalourdio:
Also, someone is maintaining an outdate `@types/intl-tel-input` in the official repository, this will lead to extreme pain in the feature

jackocnr:
> just tried 21.0.8 and does the job for me et least

👍 

> I suppose 21.0.8 will fix https://github.com/jackocnr/intl-tel-input/issues/1558 too

Unfortunately not, as that issue is about importing the react component, which has a separate types file

mpalourdio:
yes, maybe you should export react typings too. Also, this user has grabbed @types/intl-tel-input, which is incompatible

jackocnr:
> maybe you should export react typings too

Well it seems to be working for me without any explicit export configuration, as I said before, just by having the types file in the same dir as the JS file, and with the same name. This is successfully pulling in the types in my own react project.

jackocnr:
(tho I am quite new to Typescript, so I'm very open to any feedback on the setup, especially with serving 2 separate types files with this package)

mpalourdio:
I have tried with 21.0.6, and anther way is setting the `main`, as stated in the doc. With this, the `types` is not needed

`  "main": "build/js/intlTelInput.js"`

mpalourdio:
@jackocnr Also, regarding your question about multiple .d.ts, here is how `RxJs` does it for example

```package.json
"exports": {
    ".": {
      "types": "./dist/types/index.d.ts",
      "node": "./dist/cjs/index.js",
      "require": "./dist/cjs/index.js",
      "es2015": "./dist/esm/index.js",
      "default": "./dist/esm5/index.js"
    },
    "./ajax": {
      "types": "./dist/types/ajax/index.d.ts",
      "node": "./dist/cjs/ajax/index.js",
      "require": "./dist/cjs/ajax/index.js",
      "es2015": "./dist/esm/ajax/index.js",
      "default": "./dist/esm5/ajax/index.js"
    },
    "./fetch": {
      "types": "./dist/types/fetch/index.d.ts",
      "node": "./dist/cjs/fetch/index.js",
      "require": "./dist/cjs/fetch/index.js",
      "es2015": "./dist/esm/fetch/index.js",
      "default": "./dist/esm5/fetch/index.js"
    },
    "./operators": {
      "types": "./dist/types/operators/index.d.ts",
      "node": "./dist/cjs/operators/index.js",
      "require": "./dist/cjs/operators/index.js",
      "es2015": "./dist/esm/operators/index.js",
      "default": "./dist/esm5/operators/index.js"
    },
    "./testing": {
      "types": "./dist/types/testing/index.d.ts",
      "node": "./dist/cjs/testing/index.js",
      "require": "./dist/cjs/testing/index.js",
      "es2015": "./dist/esm/testing/index.js",
      "default": "./dist/esm5/testing/index.js"
    },
    "./webSocket": {
      "types": "./dist/types/webSocket/index.d.ts",
      "node": "./dist/cjs/webSocket/index.js",
      "require": "./dist/cjs/webSocket/index.js",
      "es2015": "./dist/esm/webSocket/index.js",
      "default": "./dist/esm5/webSocket/index.js"
    },
    "./internal/*": {
      "types": "./dist/types/internal/*.d.ts",
      "node": "./dist/cjs/internal/*.js",
      "require": "./dist/cjs/internal/*.js",
      "es2015": "./dist/esm/internal/*.js",
      "default": "./dist/esm5/internal/*.js"
    },
    "./package.json": "./package.json"
  },
```

jackocnr:
Thanks for your input. If we have any problems I will definitely try this out 👍

PR #1564, opened 2024-04-08
fix: Reorder exports to avoid "Error: Module not found: Error: Default condition should be last one"

mpalourdio:
While testing the very latest release, I have been caught with

`Error: Module not found: Error: Default condition should be last one`

Looks like the ordering matters. After moving, them, the error is gone.
This is so intuitive, isn't it ?

https://www.google.com/search?q=Module+not+found%3A+Error%3A+Default+condition+should+be+last+one

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,13 +95,13 @@
   "types": "./build/js/intlTelInput.d.ts",
   "exports": {
     ".": {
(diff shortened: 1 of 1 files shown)


jackocnr:
Thanks

jackocnr:
Released in v21.1.1

PR #1568, opened 2024-04-12
Area code : CANADA 879 (Newfoundland and Labrador)

florentdk59:
Hello! We noticed that CANADA 879 was missing. This is a new area code for the Newfoundland-and-Labrador province since February 17 2024. According to : https://cnac.ca/npa_codes/NPA_History.pdf and https://support.bell.ca/home_phone/what_are_the_new_area_codes_in_canada

We also hope that the webjar can be updated on maven with this new code eventually.

Thank you and kind Regards


jackocnr:
Thanks for this. Released in v21.1.4.

PR #1579, opened 2024-04-19
Adds missing exports / fixes grunt task

anthony0030:
The important files to look at are 
* grunt/translations.js
* package.json

I was testing out the code yesterday an I could not get it to work. I just realized I had a bunch of exports missing. 
These exports allowed me just to override the interface translation without touching the countries.

This is part of my code
```JS
import { enInterfaceTranslations, elInterfaceTranslations } from "intl-tel-input/build/js/i18n/index.mjs";
...

    let translations = {};
    switch (i18n.locale) {
      case "en":
        translations = enInterfaceTranslations;
        break;
      case "el":
        translations = elInterfaceTranslations;
        break;
      default:
        console.error(`${i18n.locale} is not supported in our implementation of intl_tel_input yet.`);
        translations = enInterfaceTranslations;
        break;
    }

...
        i18n: translations,
```

Idealy I would like to do this 
```JS
import { enInterfaceTranslations, elInterfaceTranslations } from "intl-tel-input/i18n";
```

instead of 
```JS
import { enInterfaceTranslations, elInterfaceTranslations } from "intl-tel-input/build/js/i18n/index.mjs";
```

Thanks 😄


build/js/i18n/ar/index.mjs
```diff
@@ -1,4 +1,7 @@
 import countryTranslations from "./countries.mjs";
 import interfaceTranslations from "./interface.mjs";
 
+export { default as countryTranslations } from "./countries.mjs";
+export { default as interfaceTranslations } from "./interface.mjs";
```
jackocnr, line 5: Am I right in thinking that lines 4&5 are duplicating the imports of lines 1&2 in this file? Would it also work to change lines 4&5 to just `export { countryTranslations, interfaceTranslations };` to avoid that duplication?

anthony0030: It may look that way but if I replace lines 1&2 with 4&5 I get errors on line 7 saying that countryTranslations is undefined.  So it is needed, I tested again this morning.

anthony0030: Line 7

```JS
export default { ...countryTranslations, ...interfaceTranslations };
```

jackocnr: Sorry I'm away from home today, so I can't check it myself but I actually meant something slightly different. I meant keep lines 1&2, and just replace lines 4&5 with that line I suggested. Do you think that would work?

anthony0030: Sorry for the confusion. I understand what you mean now. I just tested it and it works. 

This is what the code looks like
```JS
import countryTranslations from "./countries.mjs";
import interfaceTranslations from "./interface.mjs";

export { countryTranslations, interfaceTranslations };
export default { ...countryTranslations, ...interfaceTranslations };
```

It seems to be working in exactly the same way.

> Sorry I'm away from home today,...

No problem at all, whenever you get the chance. My current translation system is working. I am looking forward to making use of the inbuilt translations.

jackocnr:
Thanks so much for this Anthony 👍

Issue #1591, opened 2024-04-26
InitialValue in react component does not set the number and country state values.


KalvinHom:
### Steps to reproduce
1.  Set initialValue in the react component
2.  Do not update the country or phone number
3.  Try submitting a form with the country and number state fields
4.  number and country fields are actually blank because no onChange occurred

### Expected behaviour
number and country state fields are populated with the initialValue and its correct country.

### Actual behaviour
The variables are not populated.

### Initialisation options
initialValue="+15555555555"

jackocnr:
Good catch! Thanks for reporting.

I have a feeling the problem is just that we're initialising the plugin _before_ we start listening for the "countrychange" event (which, in turn, triggers a call to the `update` function), so I think all we need to do is switch [lines 52 and 53 in react.tsx](https://github.com/jackocnr/intl-tel-input/blob/master/react/src/intl-tel-input/react.tsx#L52-L53).

Although that relies on the plugin triggering the "countrychange" event during initialisation which I'm not sure if it does or not - if it doesn't, we'll need to add that as well.

I'm AFK at the moment, but would welcome a PR.

KalvinHom:
I took a look at this, and that didn't quite work.  

The `_setInitialState` doesn't actually call `_triggerCountryChange`.  However, when I tried adding it there, that still doesn't work because the react component won't have the `itiRef.current` set yet from having the plugin intialization complete.  The `getNumber` also won't work yet because the utils would not be finished being loaded.

Instead, I had the react component directaly call the onChange actions in its useEffect when there is an initialValue, which worked.

https://github.com/jackocnr/intl-tel-input/pull/1595

PR #1595, opened 2024-04-29
Add setting initial values in react component

KalvinHom:
Updates the react component to call the onChange for number and country when there's an initial value set.

https://github.com/jackocnr/intl-tel-input/issues/1591


jackocnr:
Thanks for putting this together. The problem with this was that it wasn't calling onChangeValidity or onChangeErrorCode. In the end I found [a slightly different solution](https://github.com/jackocnr/intl-tel-input/commit/bd29282ca86928c1e4e346f8421a05a037a325d0).

Issue #1608, opened 2024-05-10
Dark Mode Support


udit-001:
Great plugin! I love how simple it is to use.

 Are there any plans to add a dark theme to the plugin? Perhaps one that automatically adjusts based on users' browser preferences?

jackocnr:
> Great plugin! I love how simple it is to use.

❤️

> Are there any plans to add a dark theme to the plugin? 

No plans, but I would welcome a PR that added this!

jackocnr:
After playing with this a little bit, I realised 3 things that put me off the idea of providing built-in dark mode support:

- The CSS media query for dark mode `@media (prefers-color-scheme: dark)` will enable dark mode based on the user's OS system settings, BUT most websites that use this plugin won't be respecting this, and so if the plugin starts respecting this on its own, most websites will end up with a dark plugin on a light website which will look silly.
- When dark mode is enabled, the background colour for the plugin input and dropdown need to match the page background colour, which is normally some shade of dark grey, but it varies massively, and there's no easy way for the plugin to determine what that is
- It's actually really easy to add this yourself - the simplest version is just 2 CSS rules - see example below:

```css
@media (prefers-color-scheme: dark) {
  .iti {
    --iti-arrow-color: #bababa;
  }
  .iti input, .iti__dropdown-content {
    background-color: #212529;
  }
}
```

Which produces the following styles:

<img width="447" alt="Screenshot 2024-05-11 at 08 45 02" src="https://github.com/jackocnr/intl-tel-input/assets/1186883/a1885c97-25a3-4922-9818-7dc706b9058c">


So for these reasons, I've decided not to add built-in dark mode support, and instead will leave it to each dev to add this themselves, based on the above snippet, customised to match their own dark colour scheme.

PR #1614, opened 2024-05-13
Improvements to color customization

anthony0030:
Breaking changes to variable names:
--iti-text-gray -> --iti-dialcode-color
--iti-border-gray -> --iti-border-color

This pull request makes customizing colors easier and a small improvement to the search box.

This is in response to #1608, specifically:
> It's actually really easy to add this yourself
```CSS
@media (prefers-color-scheme: dark) {
  .iti {
    --iti-arrow-color: #bababa;
  }
  .iti input, .iti__dropdown-content {
    background-color: #212529;
  }
}
```

By making just those changes the result is not good. you need to customize a few more things.

if the pull request is applied you will be able to do this (with the colors you want obviously)

```CSS

@media (prefers-color-scheme: dark) {
  .iti {
    --iti-input-bg: #262a2d;
    --iti-search-bg: #262a2d;
    --iti-dropdown-bg: #262a2d;

    --iti-input-color: #ffffff;
    --iti-search-color: #ffffff;
    --iti-dialcode-color: #eeeeee;
    --iti-countryname-color: #ffffff;

    --iti-arrow-color: #ffffff;
    --iti-input-placeholder-color: #cccccc;
    --iti-search-placeholder-color: #cccccc;

    --iti-border-color: #ffffff;

    --iti-hover-color: rgba(255, 255, 255, 0.1);
  }
}
```

Because there are so many variations of dark mode, This solution makes it more customizable.

This pull doesn't have:
* A way to customize the browser's focus ring, as it would need testing on lots of browsers.
* Documentation updates


anthony0030:
Hi Jack, Thanks for reviewing this so quickly. 


> Can we tweak this so it is not a breaking change for now...

Yes, it's not a problem I understand the concern, I use ITI at work and every time there is an update I take a look at what changed, commits / issues ETC. (Because of how critical this part of code is to us)


> It's also a breaking change to specifically set the input background as white...

Yes, I get your point, I would use ```inherit``` instead of ```unset```

> Do we still need -ms-input-placeholder vendor prefixing in 2024?

🤦🏻‍♂️ My mistake I used [W3](https://www.w3schools.com/howto/howto_css_placeholder.asp) instead of MDN. Also, I forgot the project doesn't have IE/old Edge support.


> What was the reason for the placeholder opacity vars? Is this common practice?

https://dev.to/samanthaming/styling-placeholder-text-with-css-2he0
> Okay, what's this deal with opacity: 1 for Firefox. That's because, by default, Firefox's placeholders have an opacity value applied to them. So to override this, we need to set it. That way our placeholder text will show up and doesn't have the faded out appearance from the default Firefox setting.

> I see you removed the border-radius from the search input...

Very good catch. I will fix it.


> Finally, could you add a small section in the readme...

:sweat: Good documentation is like a good movie... I can't make either. (but, I will try)

I made the changes, There were some cases that I have now improved, that I would not have found if I had not made the improvements

jackocnr:
This sounds great, thanks so much, Anthony. I'm away tomorrow, but I will take another look at this on Wednesday 👍

jackocnr:
Actually, thinking about it, I don't think inherit/unset will work e.g. for input bg colour, on a page which has a bg colour other than white, the input bg colour in this case will be transparent, which is wrong (a breaking change). I just tested this with both inherit and unset.

Not sure how to handle this... needs more thought! 🤔

jackocnr:
Just spitballing here, but could you do something like this: (pseudo code):

```css
@media dark mode {
    .iti {
        background-color: var(--iti-input-bg);
    }
}
```

Without defining `--iti-input-bg` beforehand, but instead let the dev define it in their own CSS, and when they do, the only styling rule that uses it is in the dark mode media query - would that work?

anthony0030:
Yes you can reference the var before it is defined, also fyi you can default them.

> var(--custom-prop, #FF0000);

https://developer.mozilla.org/en-US/docs/Web/CSS/var


I don’t understand the problem exactly but I will try understand it better tomorrow.

could you please provide a screenshot or proof of concept to help me out.

jackocnr:
If you just take your code, and open the demo.html and change the body background colour to red, I think the input bg colour will change to red as well. Which is a breaking change - it should stay white.

anthony0030:
> [UNTESTED] I thiiiink... if you just take your code, and open the demo.html and change the body background colour to red, I think the input bg colour will change to red as well. Which is a breaking change - it should stay white.

I see the problems now. The input shows as red instead of white.

The white came from the User Agent Style Sheet, and the color was set to ```field``` So I changed it to that.
Additionally, the placeholder color has been fixed.

jackocnr:
> Good documentation is like a good movie... I can't make either. (but, I will try)

😂  you did great!

> by default, Firefox's placeholders have an opacity value applied to them

You learn a new thing every day! Good catch.

> The white came from the User Agent Style Sheet, and the color was set to field So I changed it to that

This seems to be working well 👍 

On this subject, I realised that while `inherit` makes sense for the dropdown text, it doesn't make sense for the input text, as that is just not how input text styling works normally (it doesn't inherit color from the page). So I would propose the following: we get rid of the `--iti-color` variable, change `--iti-countryname-color` to `inherit`, change `--iti-input-color` to `fieldtext` (which is the equivalent of the corresponding `field` value), and change `--iti-search-color` to `var(--iti-input-color)` - what do you think?

One other issue I found was that the new border-radius styling looks weird on Firefox and Safari - when you open the dropdown, and the search input gets the blue focus outline, it has curved corners at the top, and sharp corners on the bottom. Can we change this back to having radius on all 4 corners?

Finally, I found something really annoying! The new `--iti-input-bg:field` is actually a breaking change, because if i just have `input{background-color:yellow}` in my custom CSS, then previously the ITI input would be yellow, but now with this change, it will be white. So, I propose we leave the code exactly as it is, ready for the next breaking change, but for now, as a temporary measure, we comment out the color, bg-color, placeholder color, opacity lines for both inputs (with a comment saying "coming in v24" or whatever), and copy those 8 lines inside the dark mode media query instead - do you think that will that work?

anthony0030:
I will try to take a look at this tomorrow.

anthony0030:
I haven't been able to get to it this week.

jackocnr:
No worries at all. No rush from my end 👍

jackocnr:
Re: my last suggestion that we only apply those 8 CSS rules inside a dark mode media query - I've just realised the obvious truth that this will also be a breaking change in exactly the same way, e.g. if someone already has the following CSS:

```css
@media (prefers-color-scheme: dark) {
    input {background-color: yellow}
}
```

Then before this change, in dark mode the input would have been yellow, and after this change, in dark mode it will be white.

And presumably, for people who are interested in dark mode, this will be a common setup, to just globally change all input background colours (etc) like this. (probably not to yellow, but to a dark grey colour)

So the conclusion I'm now coming to is: we have no business breaking people's existing input styles, dark mode enabled or not, because in the vast majority of cases, this telephone input is not the only input on their site - most likely it's one of several inputs, likely on the same page, in the same form. So I think we should only expose CSS variables for them to change other plugin-specific styles e.g. the dropdown and the selected country element. What do you think?

jackocnr:
I'm really sorry to hear you've not been feeling well Anthony, I hope you're looking after yourself and resting! Get well soon.

Yeah, I think this work is really useful to show what we learnt! And also, if it's ok with you, I will leave this PR open, and I will finish it off myself when I get a moment, as I want to use most of the changes here, I will just remove the input styles, but keep all the rest.

anthony0030:
Thanks very much for your wishes and understanding.

Sorry for leaving it half-finished.

Keep up the great work! We appreciate you very much.

jackocnr:
Manually merged in v23.3.0. Thanks again Anthony! Hope you're feeling better.

anthony0030:
Hi, @jackocnr Feeling a lot better thanks!
I just tested the new version in development, 👍🏻 
There are many things going on at the moment for me, but if there is something I can help with please let me know.

jackocnr:
> Feeling a lot better thanks!

Glad to hear it

> I just tested the new version in development

Amazing, thanks. I've just put it live on https://intl-tel-input.com/ and I think it looks great - thanks again for your help.

> if there is something I can help with please let me know

❤️

PR #1665, opened 2024-06-19
Fix react `ref` Typescript types

crcarrick:
Hello 👋 

The current React types don't strongly type the forwarded ref object.  So, when using the component exported from the `intl-tel-input/react` or `intl-tel-input/reactWithUtils` packages with Typescript, you can't actually pass a `ref` to the component.

```tsx
import IntlTelInput from 'intl-tel-input/react'

// ❌ Property 'ref' does not exist on type ...
<IntlTelInput 
  ref={/* ... */}
/>
```

I added a `IntlTelInputRef` type to the `ref` object, and exported it.  The reason the type itself is exported is so that consumers can type their `useRef`, for example:

```tsx
import IntlTelInput from 'intl-tel-input/react'
import type { IntlTelInputRef } from 'intl-tel-input/react'

const ref = useRef<IntlTelInputRef>(null)

// ✅ typechecks 
<IntlTelInput 
  /* ... other props */
  ref={ref}
/>
```

No actual changes to the functionality of the React components or underlying library were made in this PR.  This is solely a Typescript fix.

### DIFF
--- a/react/build/IntlTelInput.d.ts
+++ b/react/build/IntlTelInput.d.ts
@@ -989,11 +989,11 @@ declare module "intl-tel-input/react" {
         initOptions?: SomeOptions;
         inputProps?: object;
     };
-    type ItiRef = {
+    export type IntlTelInputRef = {
         getInstance: () => Iti | null;
         getInput: () => HTMLInputElement | null;
     };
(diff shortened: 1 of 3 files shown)


jackocnr:
Looks great, thank you!

PR #1667, opened 2024-06-20
Update country names list location

emmanuel-ferdman:
# PR Summary
Small PR which adjusts the data file that contains a list of country names.


jackocnr:
Thanks

Issue #1668, opened 2024-06-22
Consistent 4:3 ratio flag images


jackocnr:
We're currently using the flag images from the [region-flags library](https://github.com/fonttools/region-flags), which has highly accurate flags, but because of this, the aspect ratios are (necessarily) inconsistent. Some flags are 15px tall e.g. Vatican City, and some are only 8px tall e.g. Qatar. And while most flags are 20px wide e.g. United Kingdom, some are only 13px wide e.g. Nepal. So if we select those 4 countries as our `onlyCountries` option, we end up with a situation like this (notice how much the flags vary in terms of width/height):

<img width="232" alt="current" src="https://github.com/jackocnr/intl-tel-input/assets/1186883/4ce36995-be0b-416f-9a90-4723d2100364">

Whereas if we switch to the [flag-icons library](https://flagicons.lipis.dev), which uses a consistent 4:3 ratio (as suggested by @bronisMateusz), we end up with this instead (notice how all flags have consistent width/height):

<img width="237" alt="new" src="https://github.com/jackocnr/intl-tel-input/assets/1186883/231ac0da-38bb-4c07-bf90-45acea33afed">

While not technically accurate in terms of respecting countries' flag aspect ratios, this provides a much more consistent UI experience.

IMPORTANT: This change will require updating/replacing the existing [image build system](https://github.com/jackocnr/intl-tel-input/blob/master/.github/CONTRIBUTING.md#updating-the-flag-images), which will become much simpler! Note: we currently use the legacy tool [Grunt](https://gruntjs.com/) for this (see the root Gruntfile.js, which triggers tasks from the root /grunt/ directory), and it would be fine to keep using that system, or it would also be fine to switch to [npm scripts](https://docs.npmjs.com/cli/v10/using-npm/scripts), which I believe would be the more modern approach.

I don't have time to implement this myself right now, but I would welcome a pull request.

jackocnr:
Released in v24.0.0

PR #1719, opened 2024-07-29
Fix: When toggle inputProps.disabled, Button for CountryList not working!

### LINKED ISSUE #1718: [BUG] When toggle inputProps.disabled, Button for CountryList not working!
Hello! Thank you for making this awesome library. I searched for several phone number libraries, and `intl-tel-input` was just perfect.

I wanted to display user's phone number with input disabled, and when the user want to modify their number, I wanted to enable input.
**But even the `inputProps.disabled` changed, a button for CountryList isn't worked (input tag was working).**

https://github.com/user-attachments/assets/d3b31306-f2c9-4e9a-8820-bfa1c22e4c7e

### Plugin version
v23.7.3

### Steps to reproduce

1. use the example code below to reproduce the problem.
```tsx
import React, { useState, ReactElement } from "react";
import { createRoot } from "react-dom/client";
import IntlTelInput from "../src/intl-tel-input/reactWithUtils";

const App = (): ReactElement => {
  const [isDisabled, setIsDisabled] = useState(true);

  const toggleDisabled = () => setIsDisabled(!isDisabled);

  return (
    <form>
      <IntlTelInput
        inputProps={{
          disabled: isDisabled
        }}
      />
      <button className="button" type="button" onClick={toggleDisabled}>Toggle</button>
    </form>
  );
};

const container = document.getElementById("app");
if (container) {
  const root = createRoot(container);
  root.render(<App />);
}
```

### Expected behaviour
Want Input and its CountryList both work when `inputProps.isDisabled` is `true`

### Actual behaviour
Even the `inputProps.isDisabled` changed from `false` to `true`, CountryList not showing! (button not working)

### Initialisation options
Used `import IntlTelInput from "../src/intl-tel-input/reactWithUtils";`

### What was the root cause?
To address the problem, I deep dived in the code and found the cause.

1. in `react.tsx`, `inputProps.disabled` passed directly into the input tag.
```tsx
// react.tsx

return (
    <input
      type="tel"
      ref={inputRef}
      onInput={update}
      defaultValue={initialValue}
      {...inputProps}
    />
  );
```

2. in `intl-tel-input.ts`, when `IntlTelInput` initialized, if `this.telInput.disabled=true` then selectedCountry disabled.

```ts
// line 840

if (this.telInput.disabled) {
          this.selectedCountry.setAttribute("disabled", "true");
}
```

3. After `inputProps.disabled` become `true`, `<input>` will be enabled because the props are directly passed (noted in 1), but selectedCountry doesn't because `intlTelInput` is already initialized. (noted in 2)

### Solution
I think user would think 'disable/enable IntlTelInput` means disable/enable both CountryList and Input tag. So I added a prop for it, and modified the code. All the tests are passed, and `allowDropdown` is also working.

https://github.com/user-attachments/assets/587ffdf1-db40-432c-96c0-f5dea6e27814

p.s. Actually this is my first time contributing open source, and I really hope to contribute in your repo. All feedbacks are welcome. Thanks 😊

bgb10:
I didn't changed the existing `inputProps.disabled` for compatibility.
Problem details are in Issue.

closes #1718


jackocnr:
Thank you so much for raising this issue, and putting together this solution! I agree that this is a bug we need to fix. But you have got me thinking now... when you said: 

> I think user would think disable/enable IntlTelInput means disable/enable both CountryList and Input tag

I think you're exactly right, and I wonder if there's a simpler and cleaner solution to achieve this goal. Looking in src/js/intl-tel-input.ts - you have highlighted the line which is the problem:

```js
// line 840

if (this.telInput.disabled) {
          this.selectedCountry.setAttribute("disabled", "true");
}
```

I wonder if just after this line, we can use a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to watch the input attributes for changes, and when the disabled attribute changes, we automatically update the selectedCountry disabled attribute  - what do you think?

bgb10:
I think there are pros and cons to each method.

# Your suggestion: use `MutationObserver` for observing disabled prop changes
pros: Just few lines to change
cons: Violation of SRP(Single Responsibility Principle) because `inputProps.disabled` manages both list and input
image
Current properties in inputProps are all for only input. But, If we use inputProps.disabled for changing 'list' and input, then it adds some ambiguity.

# My suggestion: add `disabled` prop directly in IntlTelInput (not in `inputProps`)
pros: Meets SRP and predictable for further users
cons: Compatibility Issue
There could be two way (`inputProps.disabled` and `disabled`) to change status, so it could be confusing.

> In my opinion I prefer my suggestion, so I think at first time we could use `disabled` prop directly and make `inputProps.disabled` deprecated, and then erase them at major patch (maybe v.24?)

p.s. I think it's up to you, the repo owner, to decide. If you do decide to introduce `MutationObserver`, do you mind if I PR it? I'd really like to contribute to the code, thank you!

jackocnr:
You've persuaded me! Your approach is better. The code looks perfect to me, thanks again.

Could I ask for a couple of things before I merge this:

- Update main readme with new `setDisabled` method
- Update react readme with new `disabled` prop
- Build all of the JS by running `npm run build:js` and commit that
- Add a couple of Jest tests for the new `setDisabled` method (in /tests/methods/setDisabled.test.js)

bgb10:
Oh yeah, I've done everything you suggested! Will you deprecate `inputProps.disabled` in the future? I can think of very few use cases where user want to disable only input, so I think deprecating it would be a good idea. If you do, could you please commit a deprecate mark in the code, along with the version from which it will be deprecated in readme? (it's a bit cautious for me to do that, v24~v25 should be fine in my opinion) 

Thanks :)

jackocnr:
Thanks so much for your work on this! I had just merged another PR, so there was a conflict in the build file, so I manually rebased and merged your commits. This has now been released as v23.8.0.

As for `inputProps.disabled` - I've just made the situation clear in the react readme for now. I'll consider removing it later if it causes problems.

jackocnr:
@bgb10 since you are now familiar with the project, and especially because you produced such high-quality work in this PR, I don't suppose you would have time to look at [this issue](https://github.com/jackocnr/intl-tel-input/issues/1668), which will make the flag images more consistent?

bgb10:
@jackocnr Oh thanks for suggestion. I'll do it within this week. Thanks :)

PR #1729, opened 2024-08-02
Consistent ratio flag images

bgb10:
PR for resolving #1668 

I almost done (except documentation), but there's problem in 'flag-icons' library.

Flag of Ascension Island not showing, because we use iso2 code `ac` but 'flag-icons' library uses `sh-ac`
image

Since **A**scension Island is preceded by an A in the list, I think it will stand out a lot if it was empty like it is now. 
So my suggestion is, what about giving the class name both `fi-ac` and `fi-sh-ac` when it's Ascension Island, so maybe in a future patch, we can remove it then.
image

After you give comment about this, I think the rest will be quick.


jackocnr:
Thanks for your work on this - it's great to see it coming together!

> what about giving the class name both fi-ac and fi-sh-ac when it's Ascension Island, so maybe in a future patch, we can remove it then.

Sounds good to me. So will this be a little hack in the code, something like: `if country code === 'ac'` then add this extra class? And then we'll think about clearing it up later?

Also, I notice a load of SVG files in the build dir. Are you thinking to switch from loading a single WEBP sprite image, to loading 200+ individual SVG files? I can see advantages to this approach, like they will scale up to larger display sizes easily, and maybe the browser can do some clever optimisation and only load the images that are visible as you scroll through the dropdown list (maybe?), but I would also be concerned about the massive increase in filesize for the user to download - have you considered this? I'm not saying I'm against it, I'm just saying we need to think about it and make sure it's a better experience for the user. e.g. have a look at the devtools lighthouse report etc.

bgb10:
&gt; Sounds good to me. So will this be a little hack in the code, something like: if country code === 'ac' then add this extra class? And then we'll think about clearing it up later?

1. Yes. Exactly.

&gt; but I would also be concerned about the massive increase in filesize for the user to download - have you considered this? I'm not saying I'm against it. (...) e.g. have a look at the devtools lighthouse report etc.

2. I'm really appreciated that you make me to learn what I didn't know. I did performance measurement using devtools lighthouse, and as you said, there's a significant performance bottleneck.

* Total file size: 575KiB (original) -&gt; **2703KiB (new)**
* Speed index: 0.8s (original) -&gt; **1.7s (new)**
* there's was noticable delay when user opens country dropdown at the first time (Naturally, once cached, there is no delay.)

So, I'm about to solve this like below.
1. `flag-icons` only serves `.svg` format which size is big, and they are all 'seperated' files.
2. Thus, Combining them in a single `.svg` file, and then convert it to `.webp` format which chromium browser supports optimization.
3. The size and configuration of `.webp` format will be the same as existing `flags.webp`, so we could use existing css files and class namings. (To clarify, we only need to use 'images' of `flag-icons` library)

I'll write the script for making the images. How do you think about this idea?
> Question. what is the difference between `flags.webp` and `flags@2x.webp`?

jackocnr:
> Combining them in a single .svg file (manual?), and then convert it to .webp

Firstly: please do not do this manually! That is too much work for you to do, and also it will then not be maintainable for us in the future, when flags get updated in the flag-icons project and we need to rebuild the flags.webp sprite. There will be a way to automate this, like we do currently (see grunt `img` task in Gruntfile.js - but the new approach will be simpler as every flag will be exactly the same dimensions).

Secondly: yes you could either join all the SVGs into a single SVG sprite, and then convert that single image file to WEBP, or alternatively, you could convert all the individual SVGs into WEBPs and then join those together into a single WEBP sprite. I guess whichever's easiest.

> The size and configuration of .webp format will be the same as existing flags.webp, so we could use existing css files

Unfortunately this is not true, as currently some flags have different widths e.g. see my below screenshot of the thin swiss flag in the middle:

Screenshot 2024-08-04 at 21 27 29

Whereas with the new flags, they will all have exactly the same width (and height). So you when you compose the sprite, you will need to simultaneously compose some CSS to go with it, which defines the `background-position` rules for each flag.

One final comment - can we please use the 4x3 ratio flag images instead of the 1x1? I think 4x3 is much closer to how most flags are officially supposed to be displayed.

BrianLeishman:
> One final comment - can we please use the 4x3 ratio flag images instead of the 1x1? I think 4x3 is much closer to how most flags are officially supposed to be displayed.

Just an interested party, but it would be _amazing_ if this aspect ratio was configurable somehow. We build sites for multiple clients with very different designs, and I already have two different sites where one would prefer 4:3 and one would prefer 1:1. This is actually how I found this pr in the first place. 

Awesome work btw!

bgb10:
Lately I've been too busy to do this😢  I'll continue as soon as possible.

jackocnr:
> it would be amazing if this aspect ratio was configurable somehow

I hear you, but I don't want to complicate this PR more than it already is! This is something we can consider down the line.

bgb10:
interim report: complete to write a script for multiple `.svg` to single `.webp` (also considered @1x and @2x). 
Screenshot 2024-08-10 at 4 14 01 PM

bgb10:
@jackocnr Hi Jack, I moved PR because of fussy merge conflict, so check #1750 instead.

PR #1756, opened 2024-08-22
feat: Add Polish translations for country and interface

bronisMateusz:
I need a Polish translation for my project, I prepared it for myself first, but all in all it would be nice to extend the library right away too 😁


jackocnr:
Excellent, thank you! Released in v24.2.0 💐

PR #1801, opened 2024-09-17
Refactor JavaScript Structure for Vue Components

Ibrahimmushtaq98:
Currently the structure of the Vue components feels a bit awkward to work on when trying to implement a typescript vue definition. So instead, I refactored the structure of the JavaScript related file for better organization and maintainability

Key Changes:

1. **Moved Vue JavaScript components:** The IntlTelInput.vue file and associated Vue components have been moved from vue/src/intl-tel-input/ to vue/src/js/ to make the structure more intuitive for JS-specific components.
2. **Updated Vite configurations:** The viteConfig.js and viteConfigWithUtils.js files have been relocated to vue/src/js/ to align with the new folder structure. Updated the Vite configuration to reflect the new component paths
3. Adjusted Grunt build tasks to reference the newer paths

Hopefully with these changes, it would also allows others or myself to start making Typescript vue components

Please ask question if there are any doubts!


jackocnr:
Hi there, thanks for your thoughts on this. I'm not familiar with Vue myself, so I'm very open to ideas/feedback on the setup BUT the existing directory structure was chosen because it replicates the React directory structure, and the reason for that particular organisation was precisely to get the correct module names in the (react) [typescript definition file](https://github.com/jackocnr/intl-tel-input/blob/master/react/build/IntlTelInput.d.ts). You can see how the react TS definition file is generated using [this tsconfig file](https://github.com/jackocnr/intl-tel-input/blob/master/react/tsconfig.json#L11) - specifically setting the root to the react/src directory, which means the module names are relative to that dir e.g. "intl-tel-input/react", "intl-tel-input/data" etc, which then matches how people import these modules (e.g. `import IntlTelInput from "intl-tel-input/react"`), so the types get automatically loaded and applied correctly.

Issue #1630, opened 2024-05-22
Migrate tests to Jest


jackocnr:
It's time to migrate to a modern testing setup! I've settled on Jest (and @testing-library) and I've got a few Jest tests up and running (in /tests), but there are still a lot of old Jasmine/jQuery tests to migrate over (in /src/spec). I'll keep track of the migration here. I would welcome pull requests to move more over! Thanks! 🙏 

## Core
- [ ] closeCountryDropdownEvent
- [ ] countryChangeEvent 
- [X] countrySearch
- [ ] dropdownShortcuts
- [ ] initialValues
- [ ] multipleInstances
- [ ] openCountryDropdownEvent (new)
- [ ] usingDropdown
- [ ] usingInput

## Methods
- [ ] destroy
- [ ] getExtension
- [ ] getNumber
- [ ] getNumberType
- [ ] getSelectedCountryData
- [ ] getValidationError
- [ ] isValidNumber
- [ ] isValidNumberPrecise
- [ ] setCountry
- [ ] setNumber
- [ ] setPlaceholderNumberType

## Options
- [ ] allowDropdown
- [ ] autoPlaceholder
- [ ] containerClass
- [ ] countryOrder (new)
- [ ] customPlaceholder
- [ ] dropdownContiner
- [ ] excludeCountries
- [ ] fixDropdownWidth (new)
- [ ] formatAsYouType (new)
- [ ] formatOnDisplay
- [ ] geoIpLookup
- [ ] hiddenInput
- [ ] i18n (new)
- [ ] initialCountry
- [ ] nationalMode
- [ ] onlyCountries
- [ ] placeholderNumberType
- [ ] showFlags (new)
- [ ] separateDialCode
- [X] strictMode
- [ ] useFullscreenPopup
- [ ] utilsScript
- [ ] validationNumberType (new)

## Static
- [ ] defaults
- [ ] getCountryData
- [ ] loadUtils

Mr0grog:
I know this effort is already well under way, but I had some quick feedback after starting to work on #1816 (side note: sorry for not getting to it as quickly as I’d expected). Things got a little messy when I tried to write the tests for Jest instead of the older Jasmine tests.

For #1816, I thought it was pretty important to get the actual async loading of the utils script working in the tests (it’s still a big TODO comment in Jasmine tests). For Jasmine, that was relatively straightforward — the tests just needed to run on an HTTP server instead of from a `file://` URL.

[...]

PR #1838, opened 2024-10-04
Support async functions for loading the utils script

### LINKED ISSUE #1816: Allow `utilsScript` to be a function that returns a promise for the utils module
I might be missing the obvious, but it seems like a bit of a bear to lazy load the `utilsScript` with Webpack or some other bundler, since you need to figure out how to copy the script out of the `intl-tel-input` package and keep track of the path so it can be loaded. (Please let me know if there’s a simpler way that refers to a local copy and not a public CDN!)

It would be lovely if the `utilsScript` option could a function that returns a promise for the module. That way, you could have code like:

```js
import intlTelInput from 'intl-tel-input';

const input = document.querySelector("#phone");
intlTelInput(input, {
    utilsScript: () => import('intl-tel-input/utils')
});
```

…since bundlers like Webpack can pick up on that dynamic import and split it out into a separate file automatically.

I can see that, right now, I could replace the `intlTelInput.loadUtils()` function, but that leaves me in charge of setting state-related stuff like `intlTelInput.startedLoadingUtilsScript`, which doesn’t seem like a public API I should expect to be stable.

### Plugin version

v24.5.0

Mr0grog:
Fixes #1816.

Sorry this was a bit of a long time coming! I ran into a few messy issues with Jest ([partially detailed in #1630](https://github.com/jackocnr/intl-tel-input/issues/1630#issuecomment-2392028425)) that caused me to have to do the side-note item (making sure there are no unhandled promises) to get the existing tests to run in Jest, but then I ran into complex dynamic import problems in Jest and went back to Jasmine for the new tests, and then eventually figured out the right workarounds for transforming the imports in Jest. 😩 Hopefully these commits are clearly factored enough to better evaluate the different pieces of work.

Anyway, this PR does a few main things:

1. Make sure the library never leaves promises in an unhandle-able state (thereby sometimes causing internal exceptions that can never be addressed by developers using the package). Specifically, the `loadUtils()` static function returns a promise, but when a new instance calls that function to lazy load the utils, promise rejections were never handled (the instances do get informed about the failure, but through a complex setup that leaves the failure unhandled).

    I’ve changed things to:
    - Always handle that rejection when calling `loadUtils()` internally. When users call it directly, however, it is left unhandled and they need to make sure to cover it.
    - Pass through the actual error that called the rejection. Previously, you’d just get `undefined` as the reason, and now you get a proper error object with more info, like "couldn’t import <url>” or “syntax error on line <n>”.

    This felt like the minimum change to solve the unhandle-able errors issue, although I think an ideal solution might be a bit different (see footnotes).

    I also tweaked the types for `startedLoadingAutoCountry` and `startedLoadingUtilsScript` to always be booleans (previously they were `<not-present> | true`) because I noticed some tests were broken because the test cleanup resets them to `false` instead of resetting them by deleting them (only visible as a fail if those tests run *first*). This should be pretty minor, I hope.

2. Added a server for the Jasmine tests or for viewing the `demo.html` file over HTTP, so you can test dynamic imports of the utils. (Trying to load from a `file://` URL is not allowed in most browsers for security reasons.)

    The server starts automatically for Jasmine, or you can run `npm run server` or `npx grunt jasmine:interactive` to start a long-running server on port 8000. Then you can load up `http://localhost:8000/demo.html` or `http://localhost:8000/spec.html` to debug in a browser. See changes to `CONTRIBUTING.md` for more description.

3. Added support for `loadUtils()` to take an async function that resolves to the loaded module and for the `utilsScript` option to do the same. Also renamed the option from `utilsScript` → `loadUtilsOnInit`, per [the discussion](https://github.com/jackocnr/intl-tel-input/issues/1816#issuecomment-2372350015).

    This includes expanded tests to cover actually loading (or failing to load) the code.

4. Ported the `loadUtilsOnInit` option and `loadUtils` static function tests from Jasmine → Jest.

    As part of this, I moved the Jest config into a separate file so I could add comments, but I can put it back in `package.json` if you’d prefer.

---

**Footnote**

How load failures are handled: a page with 2 inputs on it that never explicitly handles rejections of `<instance>.promise` will log 2 errors (unhandled rejections) to the browser console. That is, this setup:

```js
const instance1 = intlTelInput(inputElement1, { loadUtilsOnInit: "path/to/utils.js" });
const instance2 = intlTelInput(inputElement2, { loadUtilsOnInit: "path/to/utils.js" });
// The problem is that we don't have this code:
// instance1.promise.catch(() => doSomething());
// instance2.promise.catch(() => doSomething());
```

Will log two errors when `"path/to/utils.js"` fails to load. That’s better than it used to be, when three were logged and one of those could never be suppressed (the one coming from `loadUtils()` directly instead of from `<instance>.promise`).

However, you get the same error N times, once for each instance on the page, seems annoying. Handling all errors correctly also means developers *have* to attached handlers to `<instance>.promise`, which seems like a pain. I think a nicer solution here would be:

- Add a `intlTelInput.handleError(handlerFunction)` static method, or maybe an `"error"` event (or a differently named event that doesn’t conflict with the browser’s built-in one) that allows people to install a global error handler. If users don’t install their own handler, we’d have a default one that logs the error. But this gives people a nice, singular hook to send errors to monitoring services like Sentry/Datadog/etc. This could also handle errors from autoloading the country, etc, which has similar issues.

- Ensure `<instance>.promise` is always handled with a no-op handler. Developers can still catch rejections from it, but if they don’t, they won’t get unhandled rejection errors in their console.

This way an error loading the utils script only gets logged once, instead of for every phone input on the page, and there is a clear place to hook up and handle global errors.


grunt/replace.js
```diff
@@ -346,8 +346,8 @@ module.exports = function(grunt) {
       options: {
         patterns: [
           {
-            match: /import\(/,
-            replacement: 'import_INTENTIONALLY_BROKEN('
+            match: /import\([^)]*\)/,
+            replacement: 'Promise.reject(new Error("INTENTIONALLY BROKEN: this build of intl-tel-input includes the utilities module inline, but it has incorrectly attempted to load the utilities separately. If you are seeing this message, something is broken!"))'
```
Mr0grog, line 350: The previous approach would cause some promises to never resolve, which shouldn’t be a problem for most users, but did make tests complicated. This rejection solves that and hopefully gives users (or future contributors) a little clearer description of what’s happening.

src/js/intl-tel-input.ts
```diff
@@ -15,9 +17,9 @@ interface IntlTelInputInterface {
   getCountryData: () => Country[];
   getInstance: (input: HTMLInputElement) => Iti | null;
   instances: { [key: string]: Iti };
-  loadUtils: (path: string) => Promise<unknown> | null;
-  startedLoadingAutoCountry?: boolean;
-  startedLoadingUtilsScript?: boolean;
+  loadUtils: (source: string|UtilsLoader) => Promise<unknown> | null;
+  startedLoadingAutoCountry: boolean;
+  startedLoadingUtilsScript: boolean;
```
Mr0grog, line 22: The previous type annotation wasn’t quite correct for these two state variables (they either do not exist at all or they are `true`, never `false`). The Jasmine test suite sometimes resets them to `false` instead of deleting them, and then some tests expect them to be false, leading them to pass or fail depending on what order they were run in.

I figured it was simpler just to ensure they are always present and always a boolean.

src/js/intl-tel-input.ts
```diff
@@ -2163,6 +2211,8 @@ const intlTelInput: IntlTelInputInterface = Object.assign(
     //* A map from instance ID to instance object.
     instances: {},
     loadUtils,
+    startedLoadingUtilsScript: false,
+    startedLoadingAutoCountry: false,
```
Mr0grog, line 2215: Corresponding implementation change with the type change.

src/spec/helpers/helpers.js
```diff
@@ -149,3 +149,77 @@ var triggerKeyOnCountryContainerElement = function(key) {
...
+ * describe("my suite", function () {
+ *   useIntlTelInputBuild("build/intlTelInputWithUtils.js");
+ *
+ *   it("should do something", () => {
+ *     // In this test, window.intlTelInput will be the version from
+ *     // "build/intlTelInputWithUtils.js"
+ *   })
+ * })
+ */
+function useIntlTelInputBuild(path) {
```
Mr0grog, line 201: I left this in just in case it’s useful for something else while Jasmine tests continue to live. Can remove if it’s in the way.

You can see how this was used in 7a13ace107e7ba1f76833dcdd207e8455d782af4, before all the new tests I added that used this got removed and ported to Jest.

I did not write something similar for Jest because it wasn’t *as* important there — everything is in actual separate modules there so there isn’t global state to worry about in the same way. Having something like this could still simplify Jest tests *slightly*.

tests/helpers/matchers.js
```diff
@@ -0,0 +1,24 @@
+/** Test whether something is a promise. Works across realms. */
+function toBeAPromise(actual) {
+  const pass = Object.prototype.toString.call(actual) === "[object Promise]";
+  return {
+    pass,
+    message: () => `expected ${this.utils.printReceived(actual)}${pass ? " not" : ""} to be a promise`,
+  };
+}
```
Mr0grog, line 8: I was winding up with `Promise` instances from a different realm than the test was running in, so I couldn’t just do `expect(somePromise).toBeInstanceOf(Promise)`.

I *suspect* that the promise we get back from dynamic `import()` calls is actually from somewhere deep in Jest’s machinery, which is running in a different Node `vm` than the test itself, but am not really sure.

tests/static/loadUtils.test.js
```diff
@@ -0,0 +1,198 @@
+/**
+ * @jest-environment jsdom
+ */
+
+const intlTelInput = require("intl-tel-input");
+const { initPlugin, resetPackageAfterEach } = require("../helpers/helpers");
+require("../helpers/matchers");
+
+describe("loadUtils", function() {
```
Mr0grog, line 9: I added a little at the end of the test, but mostly this is just uncommenting and tweaking the syntax of the existing, commented-out tests.

jackocnr:
This is really incredible work. This must be the most professional and comprehensive PR I've ever received! Thank you so much for your hard work here - it is really appreciated. This is a great step forward for the project. 💐 

BTW, I agree with your footnote suggestions and would welcome a PR to that effect if you have time.

jackocnr:
Released in v24.6.0

Mr0grog:
> I agree with your footnote suggestions and would welcome a PR to that effect if you have time.

Will see if I can find some time for it later this week. 👍

jackocnr:
@Mr0grog do you know what, on second thought, I don't think it's worth adding a new static method just for the edge case of people who (A) have multiple instances, and (B) use a bad utils URL, just to prevent them from receiving a duplicate error message. The duplicate error message is really not a problem for me in this case.

I think the only thing that's important is that it is currently possible for people to catch this error if they want to. I believe this is already possible, but forgive me, I don't have time to double-check right now - can you confirm? (we should update the readme to make this clear)

Additionally, in the next major release, I plan to simplify this utils loading code so that both loadUtilsOnInit and loadUtils stop accepting a string argument and only accept the function that returns a promise which resolves to the utils module. That way, we can remove the plugin's dynamic import which will make things more straight forward.

jackocnr:
> You have to call instance.promise.catch()

Right, perfect, thanks.

> caveat the issue above if you have multiple errors, those will just be hidden from you

Ok, good to know, but I'm not worried about handling this just yet.

> FWIW, you'd still need all the same error handling and so on

👍🏻 

> and this would not have resolved most of the issues I ran into with Jest.

Sorry, I'm confused - are you saying that this proposed change would make the situation worse?

Mr0grog:
> > and this would not have resolved most of the issues I ran into with Jest.
> 
> Sorry, I'm confused - are you saying that this proposed change would make the situation worse?

No, sorry for being unclear!

It would definitely eliminate [one conditional branch from `loadUtils`](https://github.com/jackocnr/intl-tel-input/blob/05badcaeb4ecbc4e5d7a6bec508cdb868ed4b7c5/src/js/intl-tel-input.ts#L2150-L2151), and that does make for simpler code and fewer tests. 👍  But most of the complexity comes *after* turning the string into a call to `import()`, which is all stuff you’d have to keep. What I meant is that it struck me as relatively small gain for a breaking change.

As for Jest in particular, this would eliminate the need for [the `--experimental-vm-modules` option in `.npmrc`](https://github.com/jackocnr/intl-tel-input/blob/05badcaeb4ecbc4e5d7a6bec508cdb868ed4b7c5/.npmrc#L1-L3), but it would still need [the special transforms for `utils.js`](https://github.com/jackocnr/intl-tel-input/blob/05badcaeb4ecbc4e5d7a6bec508cdb868ed4b7c5/jest.config.js#L7-L23), which were really the thing that took the most work for me to figure out and get set up right. (Assuming you’d want at least one test that actually loads `utils.js` instead of a mock, so you test that the real thing does in fact work.)

Mr0grog:
That’s not to say there aren’t things you could do to simplify this funcationality. There’s some stuff I think I could have done more nicely in `loadUtils`, and I think you could probably simplify a lot of other bits (and fix some weird edge cases) by making `loadUtils()` always return a promise. You could maybe also just drop [the `removeImport` stuff from grunt-replace](https://github.com/jackocnr/intl-tel-input/blob/05badcaeb4ecbc4e5d7a6bec508cdb868ed4b7c5/grunt/replace.js#L345-L361) with relatively low impact.

PR #1839, opened 2024-10-07
Update countries.ts

FarukMZKC:
The Republic of Türkiye changed its official name from The Republic of Turkey on 26 May 2022 in a request submitted to the Secretary-General by the country's Minister of Foreign Affairs.

https://www.un.org/en/about-us/member-states/turkiye


jackocnr:
Thanks for raising this. I'm going to wait until this new name is embraced by more of the major institutions before making the change here. e.g. I notice [the BBC](https://www.bbc.co.uk/news/topics/c207p54mdq3t) still use "Turkey".

Issue #1877, opened 2024-11-20
Issues loading utils script with Angular


MarcoGlauser:
### Plugin version
 v24.7.0

### Steps to reproduce
1.  Add the `loadUtilsOnInit` option in an Angular project with an import to the options
```js 
loadUtilsOnInit: () =>  import('intl-tel-input/build/js/utils.js') 
```
2. There is no error but isValid will always be `null`

### Expected behaviour
The utils script should be loaded as expected or an error should be thrown if there was an issue.

### Actual behaviour
There is no error thrown in the console.
When attempting to load the utils script with angular, this check fails:
https://github.com/jackocnr/intl-tel-input/blob/ece711a0b8df4de458f087c2fe2f08432be49aa7/src/js/intl-tel-input.ts#L2155
My guess would be that angular patches the promise, which fails the type check and then the the utils script never gets attached to the instance. 

There was no error in the console, which made debugging this quite hard! My guess would be that this line is a bit of an overachiever:
https://github.com/jackocnr/intl-tel-input/blob/ece711a0b8df4de458f087c2fe2f08432be49aa7/src/js/intl-tel-input.ts#L922
Thankfully someone reported a similar behaviour in https://github.com/jackocnr/intl-tel-input/issues/1597 that set me on the right track.

jackocnr:
Merged, thanks again.

MarcoGlauser:
Thank you for reviewing and releasing so quickly!

PR #1878, opened 2024-11-20
Allow promise-like objects as loadUtilsOnInit option

MarcoGlauser:
Related to https://github.com/jackocnr/intl-tel-input/issues/1877
This solves the issue that angular can't load the utils script because zone.js wraps promises. 

The utils loader currently does an instanceof check against the builtin Promise type. This doesn't work with wrapped promises, bluebird promises or other custom promises.  

By wrapping the source() result in a `Promise.reslove()`, instead of the instanceof check, we're making sure that we're working with a promise.

I've added a specific testcase for a custom thenable that works with this patch.


src/js/intl-tel-input.ts
```diff
@@ -2151,10 +2151,7 @@ const loadUtils = (source: string|UtilsLoader): Promise<unknown> | null => {
       loadCall = import(/* webpackIgnore: true */ /* @vite-ignore */ source);
     } else if (typeof source === "function") {
       try {
-        loadCall = source();
-        if (!(loadCall instanceof Promise)) {
-          throw new TypeError(`The function passed to loadUtils must return a promise for the utilities module, not ${typeof loadCall}`);
-        }
+        loadCall = Promise.resolve(source());
```
Mr0grog, line 2154: @jackocnr This seems fine to me. 👍 The main thing worth considering is that this will also allow the `loadUtilsOnInit` function to return the utils module object directly instead of just as a promise. That is, you can now do:

```js
intlTelInput(inputElement, {
  // No promises involved!
  loadUtilsOnInit: () => ({
    default: { /* Locally defined utils module, or a mock, or whatever here */ }
  })
});
```

But that flexibility might be useful! So I think this is probably a good change. (Side note: the function can technically return *anything* with this change, but we have code later [line 2168] to validate that it at least vaguely looks like a JS module object, so that should be OK.)

OTOH, if you want to keep requiring that this returns a *thennable* (the more generic form of a JS promise), then the right canonical check would be something like:

```js
loadCall = source();
if (typeof loadCall?.then !== 'function') {
  // Throw an error (or maybe return a rejection instead)
}
```

…but like I said, I think the added flexibility in this change as-is is probably better.

jackocnr:
Manually resolved conflicts and merged. Thank you so much. Released in v24.8.1.

PR #1896, opened 2024-12-08
added Ukrainian translation

olekshyr:


jackocnr:
Thanks so much. Released in v25.2.0.

PR #1906, opened 2024-12-29
Improve Turkish interface translations

hasan-ozbey:
I think these translations are made by some kind of translator, they look funny.


jackocnr:
Thanks

Issue #1924, opened 2025-01-09
[FEATURE REQUEST] Add Option to Customize Country List Popup via Custom Class.


creador-dev:
#### Problem
The `intl-tel-input` library currently provides a great default UI for the country list popup. However, applying custom styles to the popup container is challenging because there is no way to scope styles exclusively to it. This can result in style conflicts that affect other parts of the application or even other plugins.

#### Proposed Solution
Add an option to specify a custom class for the popup container, similar to the existing `containerClass` option. This would allow developers to easily apply scoped styles to the popup without interfering with the library's default styles or those of other plugins.

#### Example
Introduce a configuration option, such as `popupContainerClass`, which can be passed when initializing `intl-tel-input`:

```javascript
intlTelInput(inputElement, {
  popupContainerClass: 'my-custom-popup-class',
});

jackocnr:
Hmm, why can't you just use the existing class `.iti__dropdown-content` which has the iti prefix so is likely to be unique to the plugin? Or if you need to target a specific instance of the plugin, then wrap that input in a div with a given class, or use the `containerClass` option for that instance.

creador-dev:
### Issue with Targeting Dropdown in Mobile View or Fullscreen Popup

I'm facing a challenge with styling the dropdown for `intl-tel-input` in the following scenarios:

1. When the dropdown is **ported out** (e.g., in mobile view).  
2. When `useFullscreenPopup` is set to `true`.  

In these cases, I cannot target the dropdown using `containerClass` because it doesn't wrap the dropdown with the custom class I specify.  

#### Context and Solution

I am prefixing the existing styles of `intl-tel-input` with a custom class to prevent style conflicts with other plugins that use an older version of `intl-tel-input`. Without this prefixing, the new styles from `intl-tel-input` would apply globally and mess up the styles of other plugins.

However, because the dropdown isn't wrapped in the custom class when it's ported out or in fullscreen mode, I'm unable to isolate and target it effectively.

### Goal

I need a way to ensure that the dropdown, even when ported out or in fullscreen mode, can be targeted by my custom class so that my prefixed styles apply consistently without interfering with other plugins.

jackocnr:
[...]

PR #1925, opened 2025-01-09
Add Option to Customize Country List Popup via Custom Class

creador-dev:
#### Description
This PR introduces a new `popupContainerClass` option to the `intl-tel-input` library. This option allows developers to specify a custom class for the country list popup container, making it easier to apply scoped styles and avoid conflicts with other parts of the application.

#### Why This Feature is Needed
- Custom styles can now be applied exclusively to the popup container without affecting other components or plugins.
- Helps avoid conflicts in projects with multiple instances or plugins using older versions of `intl-tel-input`.

#### Implementation
- Added `popupContainerClass` as an optional configuration property.
- Dynamically applies the custom class to the popup container during its creation.

#### Example Usage
```javascript
intlTelInput(inputElement, {
  popupContainerClass: 'my-custom-popup-class',
});


creador-dev:
Closes #1924

Issue #1928, opened 2025-01-11
[QUESTION] How to change flags size in newer version


Bartek20:
### Plugin version
latest code from github (v25.2.1)

### Question
What files do I have to modify to build flag images with changed sizes?
Documentation on website looks outdated (https://intl-tel-input.com/examples/large-flags.html).
Files described in it doesn't exist in latest source code.

* grunt/responsive_images.js - missing
* update the width/height for the retina task - which file is it
* src/css/intlTelInput.scss - it takes width, height from _metadata.scss - do I modify it here or there

Bartek20:
Also side note:
`npm run build` fails on typescript declaration generation with following error:
```
Running "shell:genTsDeclaration" (shell) task
sed: can't read : No such file or directory
Warning: Command failed: tsc --p tsconfig.json && sed -i "" -e "s/\/index\"/\"/g" build/js/intlTelInput.d.ts
sed: can't read : No such file or directory
Use --force to continue.

Aborted due to warnings.
```
And here's what happens with `npm run build -- --force` (finishes build with errors):
```
Running "shell:genTsDeclaration" (shell) task
sed: can't read : No such file or directory
Warning: Command failed: tsc --p tsconfig.json && sed -i "" -e "s/\/index\"/\"/g" build/js/intlTelInput.d.ts
sed: can't read : No such file or directory
 Used --force, continuing.

Running "shell:genReactTsDeclaration" (shell) task
sed: can't read : No such file or directory
Warning: Command failed: tsc --p react/tsconfig.json && sed -i "" -e "s/\/index\"/\"/g" react/build/IntlTelInput.d.ts
sed: can't read : No such file or directory
 Used --force, continuing.
```

Bartek20:
Ok after some searching in code commits I found that I have to change flag height in `grunt/generate-sprite.js` and that generates new scss file.

Now the problem is that error:
```
Running "generate-sprite" task
1x combined images saved as build/img/flags.webp and build/img/flags.png
Error: Error: Processed image is too large for the WebP format
    at Sharp.toFile (/workspaces/intl-tel-input/node_modules/sharp/lib/output.js:90:19)
    at createSprite (/workspaces/intl-tel-input/grunt/generate-sprite.js:162:28)
    at generateFlagMetadataAndSprite (/workspaces/intl-tel-input/grunt/generate-sprite.js:103:15)

Warning: Processed image is too large for the WebP format Use --force to continue.

Aborted due to warnings.
```
[...]

PR #1929, opened 2025-01-11
Fix some problems with sass

Bartek20:
* Fix part of vulnerabilities
* Fix deprecated sass global functions
* Moved from grunt-sass to grunt-contrib-sass


Bartek20:
Hi @jackocnr,
Sorry for pinging you. When can we expect to have it merged?

I see that you released new version since this request was created but you haven't included it in release.
Can I ask in which version will it be included.

jackocnr:
Hi there, thanks for putting this together. Unfortunately, I have very limited time to commit to open-source over the next month or so, so all I can do is merge very simple straightforward PRs e.g. basically just simple translations and LPN updates. I see you're including a translation change here, but the logic looks complicated so I'm afraid I won't be able to review this for a while.

One thing I can say is that you have included a lot of different changes in this PR, and when I have time to review it, probably in late March to be honest, I will need more information on why each change needs to happen, and why your approach fixes the issue e.g. for the Polish language change, can you describe what the current problem is, with an example, and provide a source to justify your approach here. Thanks. And apologies for the delay, but life gets in the way of open-source sometimes!

Bartek20:
@jackocnr np. I just wanted to know when merge can be expected.

Also here's why this changes need to happen:
1. Fix vulneratilities - obvious
2. Deprecated sass - deprecated function may soon be removed - [@import](https://sass-lang.com/blog/import-is-deprecated/), [map (global module)](https://sass-lang.com/documentation/modules/)
3. May no longer be needed as grunt-sass released fix after years of inactivity (no longer depends on deprecated node-sass)
4. Some languages have different form for same words in different amount of something and it can't be handled using current zero/one/more approach in counting search results<br>eg. Polish: 2 wyniki but 12 wyników and 32 wyniki<br>I solved it using `new Function()` constructor which may be considered unsafe, but input of this contructor comes directly from translations and is checked between releases so I assumed it's input is controled so should be safe.<br>Alternative approach would be to add new key to translation file (something like `searchResultFunction`) and here implement function instead of using `new Function()` constructor.
5. Inject Iti instance - Adds ability to access instance directly from DOM element (eg. `document.getElementById('phoneInput').iti`)
6. Build error - Fixed error described in [here](https://github.com/jackocnr/intl-tel-input/issues/1928#issuecomment-2585341849) and mentioned here #1941.
7. Utilities - Rn utilities are exported as `intl-tel-input/utils` but types file exports them as `intl-tel-input/utils-compiled` as it takes file name as export. I renamed file and fixed this issue (reported in #1940)

nicksleap:
Any updates?

simply-alliv:
I'm also curious on an ETA about this

simply-alliv:
Screenshot 2025-05-12 at 09 05 48

As you can see in the screenshot above, the simple but temporary fix is to disable the TypeScript error for this specific line by using a TypeScript directive comment.

jackocnr:
Hi @Bartek20, I can only apologise again for the delay in getting to this. I have had a lot come up in my personal life. But I'm here now.

Thanks again for putting all of this together. There are loads of great changes in here that I'm excited about merging! e.g. modernising the sass to get rid of the deprecation warnings, and the great catch on the utils-compiled filename issue.

Some feedback:

- Re: Polish plurals handling - thanks for bringing this issue to my attention! After looking over the code, I prefer your second idea of just putting the function straight in the translation file. This way, the pluralising logic will be more readable (especially in languages like Arabic where apparently there are 6 different kinds of plurals!), and we won't need the `new Function()` code. Are you up for making that change?
- Re: the build error fix in grunt/shell.js, I've done some research, and it seems that sed behaves differently on different platforms, e.g. on master, there's currently no error on MacOS, but if you make this change, then on MacOS it starts generating backup files that we don't need. I have found a different fix for this and pushed it to master, so please could you drop this change from this PR.
- Finally, if possible, could you rebase and drop the merge commits. Currently, when I look at the changes in this PR, it includes changes to the flag sprites, which shouldn't be there - I guess they're somehow bundled into the merge commits. Pesky merge commits! 🙃 

I'm aware it has been several months since you wrote this, so if you don't have the time/inclination to make these changes, then just let me know and I can try to do it myself. Cheers.

Bartek20:
Hi @jackocnr,
I modified translation to work from new parameter `searchResultsText`, and managed to update rest of commits to latest from your repo but it seems like I made some mistakes on the way so I had to update libphonenumber to latest in last commit and not in history (only utils were updated version stayed old). I usually only use vscode commit/push button so I don't know how to change that to be correct. I also don't know how to drop merge commits from previous updates (I only managed to merge latest changes as linear update).

jackocnr:
@Bartek20 thanks for this. The commit history on this branch is now very messy - somehow there are now 33 commits in here, including lots of old ones authored by me, and several of your own commits seem to have been duplicated as well. For this reason, I have manually squashed all of your changes into a single commit in master which you can now see [here](https://github.com/jackocnr/intl-tel-input/commits/master/). Thanks again for your great work on this!

PR #1932, opened 2025-01-15
Fix/improve strictMode for numbers with country code

### LINKED ISSUE #1931: [BUG] Strict Mode Doesn't Work When Country Code is in the Field
## Plugin Version
**v25.2.1**

## Steps to Reproduce
1. Initialize the plugin with the following options:
   ```javascript
   validationNumberTypes: [
       "FIXED_LINE",
       "MOBILE",
       "FIXED_LINE_OR_MOBILE",
       "TOLL_FREE",
       "PREMIUM_RATE",
       "SHARED_COST",
       "VOIP",
       "PERSONAL_NUMBER",
       "PAGER",
       "UAN",
       "VOICEMAIL",
   ],
   strictMode: true,
2. Add a phone number to the text field, including the country code (e.g., `+1 1234567890`).
3. Observe that the strict mode validation fails to enforce the number length.

## Expected Behavior
Strict mode should validate the phone number properly, even when a country code is present in the field. The length and type of the number should adhere to the `validationNumberTypes` configuration.

## Actual Behavior
When a country code is included in the field, the number length appears to be unlimited, and strict mode validation does not work as expected.

## Initialization Options
The following options are used for initialization:
```javascript
validationNumberTypes: [
    "FIXED_LINE",
    "MOBILE",
    "FIXED_LINE_OR_MOBILE",
    "TOLL_FREE",
    "PREMIUM_RATE",
    "SHARED_COST",
    "VOIP",
    "PERSONAL_NUMBER",
    "PAGER",
    "UAN",
    "VOICEMAIL",
],
strictMode: true,

## Additional Notes
- It seems that the country code is not being correctly factored into the validation logic when `strictMode` is enabled.
- The expected behavior should validate the full number, including the country code.

creador-dev:
Closes #1931

### DIFF
--- a/src/js/intl-tel-input.ts
+++ b/src/js/intl-tel-input.ts
@@ -1078,10 +1078,10 @@ export class Iti {
 
             // insert the new character in the right place
             const newValue = value.slice(0, this.telInput.selectionStart) + e.key + value.slice(this.telInput.selectionEnd);
-            const newFullNumber = this._getFullNumber(newValue).replace(/\D/g, "");
+            const newFullNumber = this._getFullNumber(newValue);
 			const countryCode = this.selectedCountryData.dialCode || "";
             const coreNumber = newFullNumber.startsWith(`+${countryCode}`)
-								? newFullNumber.slice(countryCode.length + 1) // Exclude "+" and country code
+								? newFullNumber.slice(countryCode.length + 1).replace(/\D/g, "") // Exclude "+" and country code
 								: intlTelInput.utils.getCoreNumber(newFullNumber, this.selectedCountryData.iso2);
             const hasExceededMaxLength = this.maxCoreNumberLength && coreNumber.length > this.maxCoreNumberLength;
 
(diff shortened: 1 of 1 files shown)


jackocnr:
Apologies for the delay in replying - life has been getting in the way.

Thanks so much for putting this together! I see you removed the code which checks if the country has changed, and I was a bit sceptical about that, so I added a few more tests around the strictMode option in master, and when I was testing this branch, I rebased it on top of master to get the latest changes, and one of the new tests is failing in this branch. Here is a link to the test that fails with your new implementation: https://github.com/jackocnr/intl-tel-input/blob/master/tests/options/strictMode.test.js#L63

I want to get this PR merged, but this test will need to pass first.

jackocnr:
I've found another solution for this: https://github.com/jackocnr/intl-tel-input/commit/d912b99d7c991659fc3eb9dce1b44d1c0133c025

PR #1939, opened 2025-01-29
Fix korean translation

moozeh:
- fixed awkward expression due to Google Translate
- unified the wording for one search result and several search results
- fixed the Ascension Island that was misinterpreted as 'Seungcheon Island(승천섬)'


jackocnr:
Thanks!

PR #1941, opened 2025-01-31
feat: add angular component

GuiEpi:
Here's my proposal for the Angular component. 

Context: I work in a company that uses intl-tel-input and we use the Angular framework. 

We propose to keep this Angular component up to date.

This addition includes:
- New Angular component
- All demos like Vue and React
- Build with grunt

I also took the liberty of correcting the `sed` command, which was giving me problems with the `-i  " "` parameter. Please let me know if this is an error on my part or not.


jackocnr:
My apologies for the delay in replying to this - life has been getting in the way.

Thank you so much for putting this together, it looks great. And I appreciate the offer of maintaining this side of the project, as I am not familiar with Angular myself.

The only thing I would say is about the empty string fix in grunt/shell.js - I looked into this a bit and found that sed behaves differently on different platforms, and the empty string argument is actually required on MacOS. So I've added a separate, platform-dependent fix for this, which is now in master. So if you could just remove that bit then I'll be happy to merge. Thanks again.

GuiEpi:
No worries, I hope all is well with you.

To tell you the truth, because of the delay I was forced to create my own lib internally.

However I would be happy to contribute here and maybe in the future we will transition from our internal lib to this one.

I added the possibility of using your component in an Angular form and I would like to add that here as well.

I'll take care of that when I have time.

jackocnr:
@GuiEpi no worries at all. I've actually gone ahead and manually made that tweak and merged this work. I've also made a couple of small updates/tweaks/fixes which should make it easier to import and use. Let me know if you have time to try it out! Thanks again for your great work!

PR #1977, opened 2025-03-10
Add aria-hidden to separate dial code element for accessibility

bchiang7:
Since the country dropdown trigger already has the dial code in it, Voiceover was announcing the dial code twice.

Before: "United States +1 +1"
With this PR: "United States +1"


jackocnr:
Apologies for the delay in getting to this, life has been getting in the way! Thanks so much for putting this together. There were some conflicts, so I merged your commit manually. Thanks again.

Issue #1966, opened 2025-02-27
Vue: Formatting is different between FF and Chrome


jpicaude:
Given the following Vue3 snippet

```
<script>
....
import { fr } from "intl-tel-input/i18n"
import IntlTelInput from "intl-tel-input/vueWithUtils"
.....
</script>

<template>
......
<IntlTelInput
    id="client-phone"
    @changeNumber="(newNumber) => console.log(newNumber)"
    :options="{
        formatAsYouType: true,
        i18n: fr,
        initialCountry: 'fr',
        nationalMode: false,
        separateDialCode: true,
    }"
    :value="form.phone"
    :modelValue="form.phone"
    class="form-control"
/>
......
</template>
```

I can see that formattting is working on Chrome but not on Firefox : 
a french number will be rendered as +33 6 12 34 56 78 on Chrome but +33612345678. 

Just cannot figure out  why...

joaovinicius:
Please, take a look

https://github.com/jackocnr/intl-tel-input/pull/1980

jackocnr:
This should be fixed in v25.4.4. Let me know if you're still having any issues.

PR #1980, opened 2025-03-21
Fix:vue component formatting on firefox

joaovinicius:
component formatting is different between vue and react  on firefox

[issue](https://github.com/jackocnr/intl-tel-input/issues/1966)

Recording 2025-03-21 at 16 33 09

Solutions:

[https://vuejs.org/guide/essentials/forms#lazy](https://vuejs.org/guide/essentials/forms#lazy)

> By default, v-model syncs the input with the data after each input event (with the exception of IME composition as [stated above](https://vuejs.org/guide/essentials/forms#vmodel-ime-tip)). You can add the lazy modifier to instead sync after change events:

Recording 2025-03-21 at 17 09 53


jackocnr:
Thanks for this work! I rebased this branch to clean up the commits and manually merged it into master. Released with v25.4.4.

Issue #1936, opened 2025-01-19
The RTL layout is not implemented correctly


mabras:
Our website uses an RTL layout, but we need the phone input to remain in LTR format.
I mean It is not good UX to put the flag/internationl prefix nubers on the right.

Current UI/UX:
![Image](https://github.com/user-attachments/assets/30bc91e6-b860-4a02-9828-8dc54d9252dd)

Expected Behavior:

![Image](https://github.com/user-attachments/assets/c4802bcd-0c84-463d-98bd-aa0baefb664f)

successbyte:
Hi,
I will start working on it please assign it to me

successbyte:
Hi, as I checked the library, it has the right to left section [here](https://intl-tel-input.com/examples/right-to-left.html) but still there is some issues:

- the country code should be **`+93`** but it is **`93+`**
- when writing the number `070 123 4567` but is shows `4567 123 070` as the placeholder

**As I work use Farsi/Persian or Arabic languages in our apps the above are real world issues**

and I will work on it if you have any comment, please share

by the way I closed above PR (I thought you do not have RTL)

![Image](https://github.com/user-attachments/assets/bf1929e1-1701-4f61-926e-51e0a5e6a420)

jackocnr:
Thanks for raising this. I've added dir=LTR to the phone input and the dial codes in the dropdown: https://intl-tel-input.com/examples/right-to-left.html

I think this issue is now fixed, but let me know if you're still having problems.

farzadaqajani:
> Thanks for raising this. I've added dir=LTR to the phone input and the dial codes in the dropdown: https://intl-tel-input.com/examples/right-to-left.html
> 
> I think this issue is now fixed, but let me know if you're still having problems.

The problem is still not solved.
The flag and separateDialCode should be on the left side of the input field and the mobile number should be written on the left side next to the flag and separateDialCode.
This problem is also visible in the demo.

<img width="1029" height="336" alt="Image" src="https://github.com/user-attachments/assets/1be04672-46e8-46ba-b58b-2168ef9f883e" />

<img width="1029" height="336" alt="Image" src="https://github.com/user-attachments/assets/9b744e56-92f6-4dce-93f2-f766778bfe4c" />

jackocnr:
[...]

PR #2000, opened 2025-04-21
GH-1936 The RTL layout issue

7snovic:
- Explicitly set the input direction instead of relying on the global dir attribute


jackocnr:
Apologies for the delay in replying - life has been getting in the way.

Thank you for putting this together. Unfortunately, I don't understand what problem this solves? Can you clearly outline what the current problem is and why this fixes it? And why do we need a new initialisation option rather than just checking for the closest RTL ancester?

jackocnr:
I've now implemented my own change to the RTL system, which actually does the opposite of this - it says if there is an RTL ancestor, then actually force the phone input and the separate-dial-code element to be LTR because phone numbers should always be LTR. So I'm going to close this for now, but if you think there is still a problem with how we handle this then please open a new issue to discuss it.

7snovic:
Hello @jackocnr Let me check your solution and if it is going to solve our issue and I will get back to you. Thank you for the efforts.

PR #2001, opened 2025-04-25
make media query for high PPI/Retina displays standards-compliant

Khuthaily:
make media query for high PPI/Retina displays standards-compliant and compatible with a wider range of browsers (forward-compatible and backward-compatible)

### DIFF
--- a/src/css/intlTelInput.scss
+++ b/src/css/intlTelInput.scss
@@ -322,7 +322,7 @@
 }
 
 //* Media query for high PPI/Retina Displays.
-@media (min-resolution: 2x) {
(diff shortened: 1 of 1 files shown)


jackocnr:
Apologies for the delay in replying.

We used to have what you are proposing, but changed it to the current implementation because it's much simpler and easier to read, and I didn't think the vendor prefix was still required anymore.

Can you provide a link to an authority that states that the vendor prefix / dpi rules are still required in 2025? I'm curious to know which browsers still require this?

jackocnr:
Closing due to inactivity.

PR #2002, opened 2025-04-26
general refactoring and improvements for CSS rules

Khuthaily:
These changes are just general improvements and refactoring for the CSS file. Here is a quick summary of the proposed changes:

- swap flags and globe images a bit faster, instead of depending on media query
- make menu scrolling smoother on modern devices
- compress padding value when necessary
- remove unnecessary specifications for input

### DIFF
--- a/build/css/intlTelInput.css
+++ b/build/css/intlTelInput.css
@@ -41,9 +41,7 @@
   overflow: hidden;
   position: absolute;
 }
-.iti input.iti__tel-input,
-.iti input.iti__tel-input[type=text],
-.iti input.iti__tel-input[type=tel] {
+.iti input.iti__tel-input {
   position: relative;
   z-index: 0;
   margin: 0 !important;
@@ -119,6 +117,7 @@
   padding: 0;
   margin: 0;
   overflow-y: scroll;
+  scroll-behavior: smooth;
   -webkit-overflow-scrolling: touch;
 }
 .iti--inline-dropdown .iti__country-list {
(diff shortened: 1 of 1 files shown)


build/css/intlTelInput.css
```diff
@@ -119,6 +117,7 @@
   padding: 0;
   margin: 0;
   overflow-y: scroll;
+  scroll-behavior: smooth;
```
lal65, line 120: This should probably be wrapped in a reduced motion user preference check.

build/css/intlTelInput.css
```diff
@@ -205,7 +204,10 @@
   width: var(--iti-flag-width);
   border-radius: 1px;
   box-shadow: 0px 0px 1px 0px #888;
-  background-image: var(--iti-path-flags-1x);
+  background-image: image-set(
```
lal65, line 207: `image-set` has [somewhat questionable support](https://caniuse.com/css-image-set).

build/css/intlTelInput.css
```diff
@@ -1188,18 +1190,12 @@
 }
 
 .iti__globe {
-  background-image: var(--iti-path-globe-1x);
+  background-image: image-set(
```
lal65, line 1193: Ditto on support for this css feature.

jackocnr:
Thanks for putting this together. I love the switch to image-set, but the only problem is, it's a breaking change, as people will be overriding the current media query to set their own custom image paths, and those overrides will no longer work if we switch to image-set. So I will add this to the list for the next major release.

As for scrolling stuff - I don't want to mess with that.

Padding values: I prefer them to be explicit.

> remove unnecessary specifications for input

This I think we need for legacy reasons - also, again, people will have their overrides and it might mess with that as well.

So I will close this, but thanks again for the image-set idea - I will copy that bit over that when the time comes! 👍🏻

PR #2018, opened 2025-05-08
Add type for inputProps

wilt-vetrec:
This PR narrows the type of `ItiProps.inputProps`from `object`, which accepts anything, to only the properties supported by the underlying `<input>` component, excluding "ref" and "onInput" which will break the component if set.


jackocnr:
Thanks for this! Merged manually, along with build files.

PR #2021, opened 2025-05-21
fix: replace the old flag of Syria with the new official flag

ARMBouhali:
Features 3 stars instead of two and different colors as per [wikipedia](https://en.wikipedia.org/wiki/Flag_of_Syria)
image

the flags@2x image was edited using Gimp then downscaled and converted to produce all 4 versions

FYI @jackocnr


jackocnr:
First, apologies for the delay in replying. Second, I'm afraid this won't work - we automatically build our flags sprite from another library, so the next time we update, these changes will be lost. Please see the section on [updating the flag images](https://github.com/jackocnr/intl-tel-input/blob/master/.github/CONTRIBUTING.md#updating-the-flag-images) in the Contributing guide.

Issue #1746, opened 2024-08-13
Feature Request - Vue 3 component


Lrochefort:
Not that I am jealous and stuff but... A nice Vue 3 component should be the next thing after the React one, since Vue.js IS the next thing after (the decline of) React...

![image](https://github.com/user-attachments/assets/ce23d55d-612a-4596-ae7b-a2eeca260b38)

jackocnr:
Haha, I completely agree with you that we should add a Vue 3 component. I have no experience with Vue, so cannot help here, but I would welcome a pull request to add this. It could be super simple to start with - just a Vue component that imports the plugin and creates an input and initialises the plugin on it.

e.g. see v1 of the react component [here](https://github.com/jackocnr/intl-tel-input/commit/794893a288e66a13ca7cd8be2fb70586f245d136#diff-2cdd1da931f80da8d9b855c8d310887a9d28d66396c4bd12c179e3c0456b2cbf) - it's literally a simple react component that renders an input, and after the first render (`useEffect`) it initialises the plugin, and calls an `update` function when things change.

You could just put it here to start with: src/js/intl-tel-input/vue.ts. And then others could help iterate from there.

Or if you were up for a slightly more complex setup, you could imitate the react setup, with a root vue/ directory, and setup symlinks to the other required plugin files (this is a workaround so that typescript module namespacing works well).

Lrochefort:
I agree. My problem is the lack of time. I may be able to do a simple PoC next weekend. To be continued…

mdpoulter:
If you're looking for a start, this is what I've used in the past (although it'll need to be converted to TypeScript):

```vue
<script setup>
import intlTelInput from 'intl-tel-input/intlTelInputWithUtils';
import 'intl-tel-input/styles';
import { onMounted, onUnmounted, ref, watch } from 'vue';

const props = defineProps({
    options: {
        type: Object,
        default: () => ({}),
    },
    disabled: {
        type: Boolean,
    },
});

const emit = defineEmits(['onChangeNumber', 'onChangeCountry', 'onChangeValidity', 'onChangeErrorCode']);

const input = ref(null);
const iti = ref(null);
const wasPreviouslyValid = ref(false);

const isValid = () => {
    if (iti.value) {
        return props.options.strictMode ? iti.value.isValidNumberPrecise() : iti.value.isValidNumber();
    }

    return null;
};

const updateValidity = () => {
    let isCurrentlyValid = isValid();

[...]

PR #2022, opened 2025-05-21
Fix changeValidity event when there is an initial value

bettysteger:
I am having trouble working with the `changeValidity` event. I would like to set the initial value when the number is empty to **true** and when not empty it should set `wasPreviouslyValid` to `isValid()`, see my change in this PR... 

Maybe @carlssonemil can take a look?

### DIFF
--- a/vue/src/intl-tel-input/IntlTelInput.vue
+++ b/vue/src/intl-tel-input/IntlTelInput.vue
@@ -86,6 +86,7 @@ onMounted(() => {
     if (props.disabled) {
       instance.value.setDisabled(props.disabled);
     }
+    wasPreviouslyValid.value = isValid();
   }
 });
 


carlssonemil:
I did not write that specific code (`wasPreviouslyValid`) so I'm not entirely sure why it is needed at all. Can't we always do the `isValid` check when the value changes and emit it, @jackocnr? I.e. remove the `wasPreviouslyValid` logic? 🤔

bettysteger:
> I did not write that specific code (`wasPreviouslyValid`) so I'm not entirely sure why it is needed at all. Can't we always do the `isValid` check when the value changes and emit it, @jackocnr? I.e. remove the `wasPreviouslyValid` logic? 🤔

@carlssonemil @jackocnr yes that would be nice :D

joaovinicius:
the same as https://github.com/jackocnr/intl-tel-input/pull/1980

jackocnr:
Hi @bettysteger @carlssonemil @joaovinicius 

Apologies for the delay in getting back to you, life has gotten in the way, but I'm here now.

Disclaimer: I have zero experience with Vue, so will leave the decision-making to you, the experts!

The initial code for this Vue component, including the `wasPreviouslyValid` bit, was written by the wonderful @mdpoulter
here: https://github.com/jackocnr/intl-tel-input/issues/1746#issuecomment-2292070527 so he's the best person to discuss that with.

Please let me know how you want me to proceed: if I should merge this, or @joaovinicius's PR #1980 or if we should try removing the `wasPreviouslyValid` code, or if we need to discuss that more first... just let me know - I'm at your service.

Edit: is this also relevant? https://github.com/jackocnr/intl-tel-input/issues/2023

carlssonemil:
If either PR (this or #1980) solves the problem I'm fine with merging either of them - or perhaps both since I don't think they will interfere with each other. Both looks good to me!

But we could also probably remove the `wasPreviouslyValid` code and do the valid check on each update instead, if @mdpoulter doesn't have a valid reason to keep it that I'm unaware of? We should discuss this first with him though, I could be missing something.

> Edit: is this also relevant? #2023

Yes, seems like it.

bettysteger:
Thanks for merging!

jackocnr:
Thanks for this work! Released in v25.4.4.

PR #2035, opened 2025-07-28
Add Country Name for Kosovo and Ascension Island.

zasadabdullah:
Add Country Name in Arabic.


jackocnr:
Thanks, but these are already in src/js/intl-tel-input/i18n/ar/interface.ts

As per the comment in that file, these 2 country names have to be in that interface file because they are not supported by the country-list library, which we use to populate country names in different languages.

PR #2042, opened 2025-08-19
Allow for importing utils during server-side rendering

mtoneil:
Currently utils can't be imported statically (included in the initial bundle instead of dynamically loaded at runtime) in a server-side rendering context because it assumes window is the global object. This PR allows for `global` as a fallback if `window` does not exist, which works in my remix app.

I made the change in closure-compiler.js and then ran `npm run build:utils` followed by `npm run build`.


jackocnr:
Thanks for putting this together, in the end I decided to add even more fallbacks (e.g. incase global doesn't exist either) which you can see here: https://github.com/jackocnr/intl-tel-input/commit/4acac69b7608a7397e8b065be7ef4471d6897b03 so I'll close this one. Thanks again for the inspiration 🙏🏻

mtoneil:
@jackocnr Thank you for the quick turnaround, much appreciated!

PR #2054, opened 2025-08-23
fix: angular import

GuiEpi:
Try to generate ts declaration with ngc instead of tsc

- Add `@angular/compiler-cli` for ngc

I hope it works! Thank you for your time.


jackocnr:
Released in v25.4.7

PR #2055, opened 2025-08-24
fix: resolve standalone component import error with ngc compilation

GuiEpi:
**Another PR that I hope will be the right one.**

I've understood that Angular needs specific metadata to recognize a standalone component. This metadata is:

```ts
static ɵcmp: i0.ɵɵComponentDeclaration<IntlTelInputComponent, "intl-tel-input", never, { ... }, { ... }, never, never, true, never>;
```

This is essential for Angular to recognize standalone components (note the `true` flag in the `ɵcmp` declaration).

### Solution
- **Removed** `"emitDeclarationOnly": true` from the TypeScript configuration
- **Used full ngc compilation** instead of declaration-only mode to preserve Angular metadata

This approach generates more files in the build directory, so please let me know if you have any concerns about this approach or would like me to adjust anything.


jackocnr:
Hi there. I'm fine with the new types dir and all the files in there. But it looks like it has duplicated all of the main build files into the angular/build dir, e.g. all of the i18n build files, which seems unnecessary - do you know why they get pulled in? If they absolutely have to be there, could we just symlink them instead of having all of those duplicates?

jackocnr:
I rebased to fix the merge conflicts and merged manually. I figured, let's just get the build working properly first, then figure out the duplicates issue 👍🏻

jackocnr:
Released in v25.4.8

GuiEpi:
> Hi there. I'm fine with the new types dir and all the files in there. But it looks like it has duplicated all of the main build files into the angular/build dir, e.g. all of the i18n build files, which seems unnecessary - do you know why they get pulled in? If they absolutely have to be there, could we just symlink them instead of having all of those duplicates?

The solution works, but I encountered an issue with the JS build because ngc doesn't transform environment variables (like `process.env.VERSION`). So now I generate a `/temp` folder with all the compiled JS files, then use esbuild to build only the essential files: `intlTelInput.js` and `intlTelInputWithUtils.js`. This approach solves the problem of creating tons of unnecessary JS files while still preserving the Angular metadata.

For the `types` folder, I believe it's fine as is and doesn't contain more than necessary or am I missing something?

jackocnr:
> The solution works

That's great news! 🎉 

> For the types folder, I believe it's fine as is and doesn't contain more than necessary or am I missing something?

Yeah, I think the types folder is fine 👍🏻

GuiEpi:
That's definitely great news! 😁  

I just did a PR with the new approach

Issue #2052, opened 2025-08-22
Way to create without wrapper, or specifying a wrapper instance?


devnix:
Hi! I'm trying to use this library with a Symfony Live Component, but I have one problem: as the form re-renders via ajax, it seems to compare the structure of the original input, his placement into the DOM, and as it doesn't find it, it creates a duplicate.

I think I could fix it by rendering a similar structure and telling intlTelInput to use my already created `iti` div as a [parentClass wrapper](https://github.com/jackocnr/intl-tel-input/blob/master/src/js/intl-tel-input.ts#L588).

Is there any actual way to accomplish this right now? If I contribute this in a pull request, would it be released soon?

Thanks!

jackocnr:
Sorry, this sounds like too much of an edge case to justify making changes to the core plugin. And I don't know anything about Symfony, so I can't help with that - I would recommend asking on StackOverflow.

devnix:
I've successfuly done It with literally 6-12 LOC. Its just adding annoptional parameter to be able to choose a wrapper instead of creating one.

I'm not aware if it's such an edge case, but it's a big blocker if you render an input from your backend, then build the intlTelInput, and you need to re-render the input. It breaks dom morphing, and also provokes rendering loops if you try to implement with Stimulus, for example. Why is having control over the wrapper such an edge case?

I can create a pull request in a moment. Im guessing that I have to build the project too and that makes the diff bigger, but its just a parameter and a little condition ☺️

PDF: I indeed tried to attempt it without modifying the core plugin, but I just did not found a way to avoid it or finding an extension point, that's why I'm proposing one

jackocnr:
Could you do this:

From your backend, send the wrapper div, so it's like this:

```html
<div class="your-backend-wrapper">
    <input />
</div>
```

Then on the frontend, when the plugin initialises, you get this:

```html
<div class="your-backend-wrapper">
    <div class="iti [other plugin classes]">
        <div class="iti__country-container">
        <input />
    </div>
</div>
```

Then, after initialisation, you immediately delete the new plugin-generated wrapper so you're back to this:

```html
<div class="your-backend-wrapper">
    <div class="iti__country-container">
    <input />
</div>
```

You could also copy over the classes from the plugin-generated wrapper to your backend-generated wrapper.

devnix:
[...]

PR #2058, opened 2025-08-26
Fix 2052

devnix:


jackocnr:
Closing, for reasons outlined here: https://github.com/jackocnr/intl-tel-input/issues/2052#issuecomment-3223944299

PR #2064, opened 2025-09-04
fix: angular component

GuiEpi:
### Fix null value handling in `writeValue`

Changed:
```ts
if (this.iti && value) {
   this.iti.setNumber(value);
}
```

to:
```ts
if (this.iti) {
   this.iti.setNumber(value || "");
}
```

This ensures `writeValue` handles null values correctly.
> Particularly useful when using `FormControl.reset()`

### Demo form component improvements

Also made some corrections to the demo form component.


jackocnr:
Released in v25.10.3. Cheers

jackocnr:
Actually, it's v25.10.4. I found that when I did a build, there were some changes in angularWithUtils, so I've pushed those and re-released.

PR #2078, opened 2025-10-06
Improve types for the placeholder options to match the implementation

Greenheart:
Adds more detailed types for the following public API options:
- `autoPlaceholder`
- `customPlaceholder`


jackocnr:
Looks good. Thanks for this. I'll merge now, but if you have time, we should also move the "off" string to the constants file, don't you think?

Greenheart:
Thanks for maintaining this widget! :)

I didn't extract `"off"` on purpose because it's never used in the implementation for the `autoPlaceholder`. It seems like you can pass any other `string` or even `null` or `0` to turn it off. However, since the current public API docs mentions `"off"`, that should be the supported value, to avoid breaking changes.

I'll create a quick PR.

The only real matches for `"off"` in the codebase (excluding duplicates from symlinks) is for the `autocomplete` attribute, which does support `"off"`.

PR #2079, opened 2025-10-06
Use placeholder constants to avoid magic strings

Greenheart:
Seems like a natural follow-up to https://github.com/jackocnr/intl-tel-input/commit/8d9e093bd2d5f2c8240b27ca8ad7e1da8e873abd and based on the comment at the top of `constants.ts`:

https://github.com/jackocnr/intl-tel-input/blob/efb4e84f962af5ac91863d4231f3371d0bac99aa/src/js/modules/constants.ts#L1-L2

Tested locally and all suites pass :heavy_check_mark:

### DIFF
--- a/src/js/intl-tel-input.ts
+++ b/src/js/intl-tel-input.ts
@@ -43,6 +43,7 @@ import {
   INPUT_TYPES,
   DIAL,
   US,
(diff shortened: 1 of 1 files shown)


jackocnr:
Makes sense. Thanks.

PR #2080, opened 2025-10-06
Use constant for `off`

Greenheart:
Follow-up to #2078


jackocnr:
Thanks for this. I actually meant to add "off" to the PLACEHOLDER_MODES. I'd rather that approach than having it separate like this and re-using it in other contexts. It was such a small change, I just pushed it myself. But thanks for all of your work!

Greenheart:
Ah, that makes sense. Make a great week! :)

Issue #2072, opened 2025-09-17
WANTED: Translators! To improve translations for countries with alternative pluralisation logic (AR, BS, CS, HR, RO, RU, SK, UK)


jackocnr:
ChatGPT tells me that the following languages (that we already support) have alternative pluralisation logic, so we should update their translation files to use the new `searchResultsText` function instead of the simple (English-based) `oneSearchResult`/`multipleSearchResults` keys:

- [x] Arabic
- [x] Bosnian
- [x] Czech
- [x] Croatian
- [X] Polish
- [x] Romanian
- [x] Russian
- [x] Slovak
- [x] Ukrainian

I would welcome pull requests from native speakers in each of these languages to make this change. See the i18n section of the readme for a simple example of how to use `searchResultsText`, or else see the [Polish translations](https://github.com/jackocnr/intl-tel-input/blob/master/src/js/intl-tel-input/i18n/pl/interface.ts#L11-L24) for a more complex example.

jackocnr:
@hasan-ozbey that's a great idea! I would welcome a pull request that moved us from the current system (importing country names from country-list project) to using DisplayNames instead.

PR #2088, opened 2025-10-23
Alternative pluralisation logic

CorvS:
This PR addresses #2072

I am not a native speaker of any of these languages and don't even understand a single word of most of them. That's why I relied on ChatGPT to tell me the correct pluralisation rules and to do the translations.

I also used other AIs (Claude, Gemini) to verify some of the rules and translations, as ChatGPT sometimes stated that there were no specific rules, then retracted that statement after I asked if it was certain.


jackocnr:
Amazing, thanks so much. I think this is a great improvement, and even if there are any small issues, this makes it much easier for any native speakers to come and suggest small fixes.

jackocnr:
Released in v25.12.1

PR #2101, opened 2025-11-01
Add instructions for running Angular demos in README

GuiEpi:


jackocnr:
Well spotted. This is my fault - since I removed all of the build files from the git repo, including the demo bundles. Thanks for this write-up.

Issue #2104, opened 2025-11-11
focus repositioning after Esc key


renow-luxembourg:
First of all, I would like to thank you for this exceptional widget.

During an accessibility audit of one of our sites, a focus repositioning issue was identified. 
Specifically, when navigating using the keyboard, if the list of countries is expanded and the Esc key is pressed, the focus is lost. The focus should be moved to the element that open the country list.

The component have to respect the RGAA/RAWeb accessibility guidelines for French-speaking countries. The test that failed is 12.8 about navigation sequence consistency (https://accessibilite.public.lu/en/raweb1/criteres.html#crit-12-8)

Note : We are currently use the version 17.0.12 of intl-tel-input but the issue is still visible on the demo page of https://intl-tel-input.com/

jackocnr:
Thanks for reporting this. And thanks for the PR as well. In the end, I decided to implement the fix in a different way - so it only triggers a focus in this specific case (Escape key), and it focuses the select-country-button instead, which is more like how native select elements work. Thanks again.

PR #2108, opened 2025-11-27
Fix focus repositioning after Esc key #2104

renow-luxembourg:
resolve: https://github.com/jackocnr/intl-tel-input/issues/2104

Hi jackocnr, this pull request includes:
- Move focus position after click on Esc key

Thanks


jackocnr:
Thanks for this. I've decided to go with a slightly different approach in the end - see my comment in the issue. Cheers.

PR #2128, opened 2026-02-01
fix: internal static reference to 'Iti' breaks when minified/mangled

rashadmehtab:
Problem: When using intl-tel-input in a production environment with aggressive minification (like Webpack's TerserPlugin or Vite's Esbuild), the class name Iti is often mangled to a single character (e.g., l). Because the constructor calls Iti._getIsAndroid(), the reference breaks if the minifier doesn't update the hardcoded Iti string inside the class logic.

This results in the error: Uncaught TypeError: l._getIsAndroid is not a function.

Solution: Refactor the call to use this.constructor or ensure the reference is handled dynamically. This ensures that even if the class is renamed during the build process, the static method remains accessible.

Affected line: this.isAndroid = Iti._getIsAndroid(); in the constructor.

### DIFF
--- a/src/js/intl-tel-input.ts
+++ b/src/js/intl-tel-input.ts
@@ -100,7 +100,7 @@ export class Iti {
     applyOptionSideEffects(this.options);
 
     this.ui = new UI(input, this.options, this.id);
-    this.isAndroid = Iti._getIsAndroid.call(this);
+    this.isAndroid = (this.constructor as typeof Iti)._getIsAndroid();
     this.promise = this._createInitPromises();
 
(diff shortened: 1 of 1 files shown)


rashadmehtab:
Updated the fix to use `(this.constructor as typeof Iti)`. This is safer for production builds because it avoids referencing the class name `Iti` directly, which can be mangled/renamed by minifiers like Terser during the build process.

jackocnr:
Thanks for raising this. In the end, I decided to extract this static class method into a separate module that is pulled in, which should also solve your issue. Released in v26.1.1.

Issue #2115, opened 2025-12-18
Copy paste broken in vue on SPA component update.


Temepest74:
I am using a basic Inertia Vue setup. I tested the latest version, and the bug still happens. From my findings, the bug was introduced in 25.10.7, as I tested versions before it, and it worked like it was supposed to.
If I copy and paste into the field, and then update the component, the content of it will be reverted to the latest version not copy-pasted. The phone number I tested is 100% valid. This only happens in strictMode: true

Small setup:
```vue
<script setup lang="ts">
import IntlTelInput from "intl-tel-input/vueWithUtils";

defineProps<{
	modelValue: string | null
	error?: string | undefined
}>();
</script>
<div>
 	<IntlTelInput
                  ref="intlTelInput"
                  :options="{
                  	strictMode: true,
                  	separateDialCode: true,
                  }"
                  :value="modelValue"
                  @change-number="emit('update:modelValue',$event)"
                  @change-validity="emit('update:validity', $event)"
        />
       <p>{{ error }}</p>
</div>
```

The form is the basic inertia useForm setup passed to the above component
```vue
form = useForm<FormDataType<FormData>>({
	"phone": null,
});
```

This bug doesn't occur if no error is set into the error prop. One thing to note is that, if I resubmit the form after an error (so the field looks empty / has the old value in it), it submits the copy-pasted value anyway, so this is just a visual thing.

jackocnr:
I'm not familiar with Vue myself, but maybe @carlssonemil can help? 🙏🏻

carlssonemil:
> I'm not familiar with Vue myself, but maybe [@carlssonemil](https://github.com/carlssonemil) can help? 🙏🏻

Sorry for the late reply, been away on vacation 🌴 

I *think* I know what the problem is and will try to find the time to confirm my suspicion. Essentially I think we need to add the `@paste` event on the input element, like so:

```vue
<input
    ref="input"
    v-model.lazy="model"
    type="tel"
    @countrychange="updateCountry"
    @input="updateValue"
    @paste="updateValue"
    v-bind="inputProps"
  />
```

jackocnr:
Hi @carlssonemil, I was asking AI about this problem, and it recommended the following:

[...]

PR #2130, opened 2026-02-02
fix: Paste reverting input value on component rerender (#2115)

carlssonemil:
### Summary
Fixes an issue (#2115) where pasted phone numbers could visually revert after a component re-render (e.g. on validation errors) when `strictMode` is enabled for the included Vue-component.

### Root cause

The input value, Vue state, and `intl-tel-input` instance could become desynchronized. When the parent rerendered (e.g. due to error updates), Vue would overwrite the DOM value with stale state, causing the pasted value to appear reverted even though the correct value was submitted.

### Changes

- Make the parent-controlled value (i.e. the external value) the single source of truth.
- Always emit canonical values from the `intl-tel-input` instance.
- Re-sync plugin state after rerenders to prevent visual rollback.
- Remove lazy / duplicate model updates and improve null safety.

### Additional changes

- Add support for both `modelValue` and `value` props to enhance compatibility with different usage patterns.
- Update Vue demos with correct imports.

### DIFF
--- a/vue/src/IntlTelInputWithUtils.vue
+++ b/vue/src/IntlTelInputWithUtils.vue
@@ -2,40 +2,37 @@
 <script setup lang="ts">
 import intlTelInput from "./intl-tel-input/intlTelInputWithUtils";
 import type { SomeOptions } from "./modules/types/public-api";
-import { onMounted, onUnmounted, ref, watch } from "vue";
+import { onMounted, onUnmounted, ref, watch, computed } from "vue";
 import type { InputHTMLAttributes } from "vue";
 
-const modelValue = defineModel<string>({
-  default: ''
-});
-
 interface Props {
-  options?: SomeOptions,
-  usePreciseValidation?: boolean,
-  disabled?: boolean,
-  inputProps?: InputHTMLAttributes
+  options?: SomeOptions;
+  usePreciseValidation?: boolean;
+  disabled?: boolean;
+  inputProps?: InputHTMLAttributes;
 }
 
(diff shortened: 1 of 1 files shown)


jackocnr:
Hi @carlssonemil thank you so much for putting this together! Hope you had a great trip!

> Make the parent-controlled value (i.e. the external value) the single source of truth.

Help me understand this bit. Is this more like a controlled input or uncontrolled input?

carlssonemil:
> Hi @carlssonemil thank you so much for putting this together! Hope you had a great trip!

I did, thanks for asking! It was very much needed 😁 

> > Make the parent-controlled value (i.e. the external value) the single source of truth.
> 
> Help me understand this bit. Is this more like a controlled input or uncontrolled input?

It's just to ensure that the external value coming through the `value`/`modelValue` prop is the one we always use internally in the Vue-component and the one we apply to the `intl-tel-input` instance. Mainly to reduce any potential desynchronization between the internal value and the value that is emitted up and visible to the end-user. Before these changes we didn't always update the internal value, hence why the pasted number was visibly reverted on rerenders.

It's mainly this watcher doing the work:
```javascript
watch(
  () => displayed.value,
  (val) => {
    if (!instance.value) return;

    // Avoid cursor jumping when typing
    const next = val ?? "";
    const currentCanonical = instance.value.getNumber?.() ?? "";

    if (currentCanonical === next) return;

    instance.value.setNumber(next);
    updateValidity();
  },
  { flush: "post" },
)
```

jackocnr:
Amazing, thanks again 💐

PR #2136, opened 2026-02-18
fix(angular): mark phone control as touched on submit

GuiEpi:
- Add `markAsTouched` when submit form


jackocnr:
Thanks Guillaume!

jackocnr:
Hmm actually it's not working for me. I go to the angular component example page and click submit and nothing happens 🤔

PR #2138, opened 2026-02-19
docs: fix angular component

GuiEpi:
## Fix Angular component example styling

Added visual feedback for the disabled submit button:
- Reduced opacity when form is invalid
- Changed cursor to `not-allowed` for better UX

Also removed unnecessary `markAsTouched()` call since the button is disabled when the form is invalid anyway (I had forgotten that I disabled the button in this example to showcase Angular reactive forms "power").

---

Sorry for the noise with the previous PRs...

### DIFF
--- a/angular/demo/form/form.component.ts
+++ b/angular/demo/form/form.component.ts
@@ -14,7 +14,13 @@ import { IntlTelInputComponent } from '../../src/intl-tel-input/angularWithUtils
           initialCountry: 'us',
         }"
       />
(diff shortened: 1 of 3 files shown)


jackocnr:
Thanks for this. In the end, I decided to go with a different approach (removing the disabled state from the button). Previously, the validate button was disabled until there was a valid number, but this actually hides the error message functionality, which I want this demo to show off. This brings it in line with the React example. Thanks again.

PR #2144, opened 2026-02-28
fix(angular): improve form validation UX by resetting submit state on input change

GuiEpi:
### Problem
When using the Angular component with reactive forms, there was a UX issue where:
1. User enters a valid phone number (e.g., `06 63 77 26 63`)
2. User clicks "Validate" → displays `Valid number: +33663772663`
3. User modifies the number (e.g., changes last digit from 3 to 8)
4. The success message still shows the **old number** `+33663772663` instead of the new one
5. User must submit again to see the updated number

This created confusion because:
- The displayed number didn't match the actual input
- The "Validate" button seemed to lose its purpose after the first click

### Solution
- **Reset validation state on change**: When the user modifies the phone input, reset the `isSubmitted` flag so the success message disappears
- **Clear UX flow**: User modifies → success message disappears → user validates again → see correct formatted number

### Changes
- Updated `angular/demo/form/form.component.ts` (demo component)
- Updated `site/src/examples/js/angular_component.ts.ejs` (functional example on website)
- Updated `site/src/examples/js/angular_component_display_code.ts` (code displayed to developers)

### DIFF
--- a/angular/demo/form/form.component.ts
+++ b/angular/demo/form/form.component.ts
@@ -1,4 +1,4 @@
-import { Component, ViewChild } from '@angular/core';
+import { Component, OnInit, ViewChild } from '@angular/core';
 import { FormControl, FormGroup, ReactiveFormsModule, Validators } from '@angular/forms';
 import IntlTelInput from '../../src/intl-tel-input/angularWithUtils';
 
@@ -20,8 +20,8 @@ import IntlTelInput from '../../src/intl-tel-input/angularWithUtils';
           Phone number is required.
         } @else if (phone?.errors?.["invalidPhone"] && phone?.touched) {
           {{ phone?.errors?.["invalidPhone"].errorMessage }}
-        } @else if (notice) {
-          {{ notice }}
+        } @else if (isSubmitted && fg.valid) {
+          Valid number: {{ telInput.getInstance()?.getNumber() }}
         }
       </div>
     </form>
@@ -29,23 +29,29 @@ import IntlTelInput from '../../src/intl-tel-input/angularWithUtils';
   standalone: true,
   imports: [IntlTelInput, ReactiveFormsModule]
 })
-export class AppComponent {
+export class AppComponent implements OnInit {
   @ViewChild('telInput') telInput!: IntlTelInput;
 
   fg: FormGroup = new FormGroup({
     phone: new FormControl<string>('', [Validators.required]),
   });
 
-  notice: string | null = null;
+  isSubmitted = false;
 
   get phone() {
     return this.fg.get('phone');
   }
 
+  ngOnInit(): void {
+    this.phone?.valueChanges.subscribe(() => {
+      this.isSubmitted = false;
+    });
+  }
+
   handleSubmit(): void {
     this.phone?.markAsTouched();
     if (this.fg.valid) {
-      this.notice = `Valid number: ${this.telInput.getInstance()?.getNumber()}`;
+      this.isSubmitted = true;
     }
   }
 }
--- a/site/src/examples/js/angular_component.ts.ejs
+++ b/site/src/examples/js/angular_component.ts.ejs
@@ -1,7 +1,7 @@
 import "zone.js";
 import "@angular/compiler";
 import { bootstrapApplication } from "@angular/platform-browser";
-import { Component, ViewChild } from "@angular/core";
+import { Component, OnInit, ViewChild } from "@angular/core";
 import { FormControl, FormGroup, ReactiveFormsModule, Validators } from "@angular/forms";
 import IntlTelInput from "../../../build/intl-tel-input/angular/IntlTelInput.js";
 
@@ -30,8 +30,8 @@ import IntlTelInput from "../../../build/intl-tel-input/angular/IntlTelInput.js"
           Please enter a number
         } @else if (phone?.errors?.["invalidPhone"] && phone?.touched) {
           {{ phone?.errors?.["invalidPhone"].errorMessage }}
-        } @else if (notice) {
-          {{ notice }}
+        } @else if (isSubmitted && fg.valid) {
+          Valid number: {{ telInput.getInstance()?.getNumber() }}
         }
       </div>
     </form>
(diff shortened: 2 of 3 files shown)


jackocnr:
Well spotted! Thanks for putting this together. I have reviewed this proposed solution, and thought about it, and for now, I'd like to keep the functionality in line with the other component example pages, so it should show the new valid message automatically when a new valid number is typed.

I ended up using AI to generate an alternative solution, which also centralises all the notice logic in one place away from the template, which feels cleaner to me, and again, brings it in line with the other component examples. Let me know what you think: [site: fix angular example bug by centralising notice logic](https://github.com/jackocnr/intl-tel-input/commit/437dccc3255706e916a1b47b3bcc9b41837cd61b).

GuiEpi:
I think your implementation is very good, but I have two comments.

1. I have exposed a constant `PHONE_ERROR_MESSAGES` to avoid having to create `errorMap` each time.
2. I implemented my example this way because I find that the fact that the "number changes on its own" makes the validate button lose its meaning, and it is not really clear what it is for except on the first click. But that's not a real problem and your implementation works perfectly.

Edit: Forget about my `PHONE_ERROR_MESSAGES` I just realised why you created `errorMap`. Perhaps my idea to create this constant was silly?

jackocnr:
Thanks for your feedback 🙏🏻 

Re: PHONE_ERROR_MESSAGES, I like the idea, but the downside to hardcoding error messages in the component like this is (1) it doesn't allow for devs to customise these strings to their own wording style, to be consistent with the rest of their site, and (2) it doesn't allow for localisation. So I would vote to drop this.

Re: the validate button becomes useless after the first click: yes, I see your point, and you are completely right. I have thought about this and decided your proposed solution makes more sense. But for some reason, your PR seems to have all the v27 commits in - perhaps you applied it on top of the v27 branch? Would you be able to apply it on top of master instead, and submit a new PR? This is a change that we should be able to launch immediately, without waiting for the v27 branch to be merged (which might not happen for a while).

I have since implemented this approach on the [React example page](https://intl-tel-input.com/examples/react-component) along with Bootstrap's validation styling, and I think that it works really well now - what do you think? Would you be up for adding the Bootstrap validation styling as well? You just have to add `novalidate` attribute to the form, and then add "is-valid" or "is-invalid" class to the tel input as needed, and it should magically work!

GuiEpi:
> I have since implemented this approach on the [React example page](https://intl-tel-input.com/examples/react-component) along with Bootstrap's validation styling, and I think that it works really well now - what do you think? Would you be up for adding the Bootstrap validation styling as well? You just have to add novalidate attribute to the form, and then add "is-valid" or "is-invalid" class to the tel input as needed, and it should magically work!

It looks great, I'll take care of implementing it! I will also remove my constant.

> But for some reason, your PR seems to have all the v27 commits in - perhaps you applied it on top of the v27 branch? Would you be able to apply it on top of master instead, and submit a new PR?

Sorry, I actually thought I had to add my code to v27, but I'll redo it directly on master.

jackocnr:
Btw, I did briefly try asking AI to implement this change, and the implementation got a bit messy (and a bit out of my depth) because it identified a problem where the inputProps (specifically the "class" field) was only being applied once, at initialisation time, and then any changes after that were ignored. So after initialisation, the user types their number and the validation state changes, it tries to update the "is-valid" (etc) class in inputProps, but this change doesn't have any effect (meaning it doesn't update the DOM). So I think that in order to get this to work, the underlying Angular wrapper component will have to be updated to watch inputProps for changes and reapply them whenever that happens. Does that make sense?

jackocnr:
Further to this, AI came up with the following solution - let me know your thoughts: https://github.com/jackocnr/intl-tel-input/commit/1d752572352f3b7e44d2c8c17661115a35b07e38 (this is in a branch called `angular-fix` if you want to have a play)

jackocnr:
Hi @GuiEpi, I've gone ahead and merged that commit into master, and it's now live on the demo site, but please do let me know if it looks ok to you or if you'd rather take a different approach. I won't do a new release until I hear from you.

GuiEpi:
Hi, @jackocnr 

Sorry, I'll look into it very soon. Yesterday and today were busy days at work, but things should calm down tomorrow.

I noticed a few things that I think could be improved. 

In addition, there is someone who previously maintained an [npm package that depended on intl-tel-input](https://www.npmjs.com/package/ngx-intl-tel-input) and integrated Bootstrap. I think there is some good stuff in that (he must have encountered the problem you mentioned above). I will take a look at it.

I will try to do it quickly.

jackocnr:
Hi @GuiEpi 

I got the Claude Max for open source! I asked it for feedback on the Angluar component, and it gave the following, in case this is helpful for you:

**Overall**: Well-structured wrapper with good ControlValueAccessor/Validator integration. A few issues worth addressing:

---

### Bugs / Correctness

1. **Initialization in `ngOnInit` instead of `ngAfterViewInit`** (line 126-137): You initialize `intlTelInput` in `ngOnInit`, then set `initialValue` in `ngAfterViewInit`. This works because of `{ static: true }` on the ViewChild, but it splits related initialization logic across two lifecycle hooks unnecessarily. Moving it all to `ngAfterViewInit` would be clearer and more conventional.

2. **`writeValue` doesn't trigger change propagation** (line 303-307): When a reactive form patches the value via `writeValue`, you call `setNumber` but never call `handleInput()`. This means `numberChange`, `validityChange`, `countryChange`, and `errorCodeChange` won't emit, and the validator state won't update until the user types.

3. **`errorCode` can be out of bounds for `PHONE_ERROR_MESSAGES`** (line 338): `getValidationError()` could return a value not covered by the 5-element array, resulting in `undefined` in the error object. A fallback like `PHONE_ERROR_MESSAGES[errorCode] ?? "unknown"` would be safer.

---

### Design

5. **`inputProps` typed as `object`** (line 94): This provides no type safety. A typed interface (with known HTML input attributes) would give consumers better autocomplete and catch typos.

6. **Output names shadow native events** (lines 102-107): `blur`, `focus`, `keydown`, `keyup`, `paste`, `click` shadow the native DOM event names. This can confuse consumers about whether they're getting the wrapper's event or the native one. Consider prefixing (e.g., `inputBlur`) or documenting this clearly.

7. **`hasChanged` flag is redundant** (lines 165-199): You track `hasChanged` to conditionally call `onValidatorChange()`, but you could just always call it — Angular's forms infrastructure handles no-op validation efficiently, and it simplifies the logic.

---

### Minor

8. **Three identical eslint-disable comments** (lines 119-124): Could use a single block disable or configure the rule to allow this pattern for CVA implementations.

9. **`getInput()` never returns null** (line 239-241): The return type says `| null` but `nativeElement` is always defined after `static: true`. The signature is misleading.

PR #2145, opened 2026-03-07
refactor(angular): simplify Angular input props and lifecycle

GuiEpi:
I started again from my "basic component" which is on the last tag.

- Remove any extra attributes that may be added via `inputProps`
- Remove `ngOnInit`, keep only `ngAfterViewInit`
- Other minor improvements to the types
- Handle the case -99 `UNKNOWN_VALIDATION_ERROR`
- Form demo: improve form validation UX

Removed

> writeValue doesn't trigger change propagation (line 303-307): When a reactive form patches the value via writeValue, you call setNumber but never call handleInput(). This means numberChange, validityChange, countryChange, and errorCodeChange won't emit, and the validator state won't update until the user types.

The `writeValue` behavior is correct as-is. In Angular, `writeValue` is a one-way operation: the form model pushes a value into the component's view, but the component should NOT emit change events back. This is how Angular's own built-in `<input>` works too — if you call `formControl.setValue("hello")`, the input's value updates in the DOM, but no `(input)` or `(change)` event fires.
e.g. https://github.com/angular/angular/blob/main/packages/forms/src/directives/default_value_accessor.ts#L116-L123

> errorCode can be out of bounds for PHONE_ERROR_MESSAGES (line 338): getValidationError() could return a value not covered by the 5-element array, resulting in undefined in the error object. A fallback like PHONE_ERROR_MESSAGES[errorCode] ?? "unknown" would be safer.

I made a fix to handle the case -99 `UNKNOWN_VALIDATION_ERROR`.

> inputProps typed as object (line 94): This provides no type safety. A typed interface (with known HTML input attributes) would give consumers better autocomplete and catch typos.

Change it to `Record<string, string>`. It will be clearer in v27 when this function is renamed to `inputAttributes`

> Output names shadow native events (lines 102-107): blur, focus, keydown, keyup, paste, click shadow the native DOM event names. This can confuse consumers about whether they're getting the wrapper's event or the native one. Consider prefixing (e.g., inputBlur) or documenting this clearly.

Using the native event names (`blur`, `focus`, etc.) is intentional. It matches what Angular developers expect. Prefixing would hurt the developer experience for no real benefit.

> hasChanged flag is redundant (lines 165-199): You track hasChanged to conditionally call onValidatorChange(), but you could just always call it — Angular's forms infrastructure handles no-op validation efficiently, and it simplifies the logic.

The `hasChanged` flag and deduplication logic is actually your implementation, and I think it's solid. It prevents emitting duplicate events to the parent on every keystroke, which avoids unnecessary change detection cycles. I'd keep it as-is.

> Three identical eslint-disable comments (lines 119-124): Could use a single block disable or configure the rule to allow this pattern for CVA implementations.

Maybe "configure the rule to allow this pattern for CVA implementations" ?

> getInput() never returns null (line 239-241): The return type says | null but nativeElement is always defined after static: true. The signature is misleading

I think that's a good point. I'll update the return type to `HTMLInputElement` in v27 to avoid the unnecessary null check for consumers, since it's technically a breaking change (I think).

---

If there are some points you would like to discuss, do not hesitate.

### DIFF
--- a/angular/src/intl-tel-input/angular.ts
+++ b/angular/src/intl-tel-input/angular.ts
@@ -76,7 +76,7 @@ export class IntlTelInputComponent
 
   @Input() initialValue?: string;
   @Input() usePreciseValidation: boolean = false;
-  @Input() inputProps: object = {};
+  @Input() inputProps: Record<string, string> = {};
   @Input() disabled: boolean = false;
   @Input() initOptions?: SomeOptions;
 


jackocnr:
This looks grand, thanks so much @GuiEpi. Amazing that you found a way to get inputProps.class (etc) to automatically update without adding all those individual lines to the input element, like in the previous (AI) implementation. Great simplification. I think in v27, we should explicitly strip "disabled" and "readonly" and maybe other keys from inputAttributes to enforce our way of handling those situations, but yeah it makes sense to leave that to v27. Thanks again.

GuiEpi:
In reality, your AI implementation had corrected the error using this piece of code:
```ts
if (changes["inputProps"]) {
  this.applyInputProps();
}
```
For the rest, I think it "over-engineered" the solution.

> I think in v27, we should explicitly strip "disabled" and "readonly" and maybe other keys from inputAttributes to enforce our way of handling those situations, but yeah it makes sense to leave that to v27.

I'm not really sure what to do. I think that if I expose these props, they will inevitably end up in the DOM. It might be better to let the user do it via `inputAttributes`. That would avoid always having `readonly=""` or `placeholder=""` in the DOM.

---

This is unrelated, but I also tried to get Claude Max for open source, and I am pleased to tell you that they accepted! 🎉🎉

jackocnr:
> I'm not really sure what to do. I think that if I expose these props, they will inevitably end up in the DOM. It might be better to let the user do it via inputAttributes. That would avoid always having readonly="" or placeholder="" in the DOM.

I'm confused. Currently, they can pass "disabled" in inputProps, right? But setting that attribute directly can leave the plugin in an inconsistent state, as the country dropdown can still be clickable. So instead we should strip/ignore that key from inputProps (soon to be inputAttributes), and so force them to use the separate "disabled" prop instead, right? (and do the same for readonly)

> This is unrelated, but I also tried to get Claude Max for open source, and I am pleased to tell you that they accepted!

Amazing! Congratulations! ❤️

PR #2148, opened 2026-03-14
feat: add clearOnCountryChange and clearOnDropdownClose options + reset() method

### LINKED ISSUE #2146: Clear search input and reset filtered results after country change
I’m wondering if there is a way to clear or reset the value in the search input after the selected country changes.

Currently, I’m using the following workaround:
```
input.addEventListener("countrychange", () => {
    $(".iti__search-input").val("");
});
```
However, this is not an ideal solution because it only clears the search input field. The dropdown results remain filtered based on the previous search, so the country list is not fully reset.

It might be helpful to provide configuration options such as:
```clearOnCountryChange: true```
```clearOnDropdownClose: true```

Methods:
```itiInstance.reset()```
```itiInstance.clear()```

### PR DESCRIPTION
**Pull request by benriemer**
Fixes #2146

- Add `clearOnCountryChange` option to clear search input when country changes
- Add `clearOnDropdownClose` option to clear search when dropdown closes
- Add public `reset()` method to manually clear search and filtered results

Usage:
```js
// Auto-clear when country changes
const iti = intlTelInput(input, { clearOnCountryChange: true });

// Auto-clear when dropdown closes  
const iti = intlTelInput(input, { clearOnDropdownClose: true });

// Manual reset
iti.reset();
```

### DIFF
--- a/src/js/intl-tel-input.ts
+++ b/src/js/intl-tel-input.ts
@@ -1190,6 +1190,11 @@ export class Iti {
     // Delegate DOM-only close behaviour to UI
     this.#ui.closeDropdown();
 
+    // Clear search input on dropdown close if option is enabled
+    if (this.#options.clearOnDropdownClose) {
+      this.#ui.handleSearchClear();
+    }
+
     //* Unbind dropdown-scoped events in one go
     this.#dropdownAbortController.abort();
     this.#dropdownAbortController = null;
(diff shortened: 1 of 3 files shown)


jackocnr:
Thanks for putting this together, but as I said in the issue thread, I will just change the default behaviour in v27 - I'd rather not add more initialisation options (we already have 28 options!).

PR #2155, opened 2026-03-22
fix: add types field to ./data and ./utils exports for moduleResolution: bundler

### LINKED ISSUE #195: Country code select tab navigation

benriemer:
## Summary

Fixes TypeScript moduleResolution: "bundler" error (issue #195)

## Problem
When using `moduleResolution: "bundler"` (standard in Vite/Next.js), imports from `intl-tel-input` fail because the `./data` and `./utils` exports were missing the explicit `types` field in package.json exports.

## Solution
Add explicit `types` field to `./data` and `./utils` exports, pointing to the bundled type definitions in `./build/js/intlTelInput.d.ts`.

## Changes
- `package.json`: Updated exports for `./data` and `./utils` to include types field

## Testing
The fix has been verified to provide valid type information for bundler resolution.

---
Closes #195

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,8 +125,16 @@
       "import": "./build/js/intlTelInputWithUtils.js",
       "default": "./build/js/intlTelInputWithUtils.js"
     },
-    "./data": "./build/js/data.js",
-    "./utils": "./build/js/utils.js",
+    "./data": {
+      "types": "./build/js/intlTelInput.d.ts",
+      "import": "./build/js/data.js",
+      "default": "./build/js/data.js"
+    },
+    "./utils": {
+      "types": "./build/js/intlTelInput.d.ts",
+      "import": "./build/js/utils.js",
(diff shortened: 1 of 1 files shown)


jackocnr:
Thanks for this. I don't understand how it's connected to the referenced issue (I suspect this is AI gloop?), but the code change looks good. You included an irrelevant commit in this PR, so I just manually cherry-picked the good commit and pushed it to master, so that will be included in the next release. Thanks again.

jackocnr:
Released as part of v26.9.0. Thanks again.

PR #2156, opened 2026-03-24
fix(TypeError): guard `selectedCountry` before toggling disabled

### LINKED ISSUE #2154: [Angular] Updating `disabled` prop throws: Cannot read properties of undefined (reading 'setAttribute')
### Prerequisites

- [x] I am using the latest version (v26.8.1) of both intl-tel-input and utils.js.
- [x] This plugin uses libphonenumber for formatting, validation, and placeholder numbers. If my issue relates to one of these things, I confirm I have used their [test site](https://libphonenumber.appspot.com) to ensure the issue is not with them, and will provide a link to the test results page showing this. Else, I have [reported the issue](https://github.com/google/libphonenumber/blob/master/CONTRIBUTING.md) to them instead of here.

### Current behaviour

When providing this config:

```ts
{
  strictMode: true,
  nationalMode: true,
  separateDialCode: false,
  allowDropdown: false,
  showFlags: false,
  initialCountry: 'FR',
  onlyCountries: ['FR'],
  containerClass: 'w-100', // Should not be relevant for the repro steps
  searchInputClass: 'form-control w-100', // Should not be relevant for the repro steps
} 
```

While switching the `disabled` option bound to a [Signal](https://angular.dev/guide/signals) it throws `Cannot read properties of undefined (reading 'setAttribute')` in the console.

Image

It seems to happen because in the [setDisabled](https://github.com/jackocnr/intl-tel-input/blob/master/src/js/intl-tel-input.ts#L1626) method the `this.#ui.selectedCountry` element is being accessed but it's not available on the private `ui` member because (I assume) the `allowDropdown` has been disabled.

A possible fix for this would be:

```ts
  public setDisabled(disabled: boolean): void {
    if (!this.#ui.telInput) {
      return;
    }
    this.#ui.telInput.disabled = disabled;
    if (disabled) {
      this.#ui.selectedCountry?.setAttribute("disabled", "true");
    } else {
      this.#ui.selectedCountry?.removeAttribute("disabled");
    }
  }
```

### Expected behaviour

Switching the `disabled` prop should not throw an error.

### Steps to reproduce

1. Use this config
```ts
{
  strictMode: true,
  nationalMode: true,
  separateDialCode: false,
  allowDropdown: false,
  showFlags: false,
  initialCountry: 'FR',
  onlyCountries: ['FR'],
  containerClass: 'w-100', // Should not be relevant for the repro steps
  searchInputClass: 'form-control w-100', // Should not be relevant for the repro steps
} 
```

2. Bind a bool signal to the `disabled` prop
3. Switch the signal from true to false (or the other way around)

### Anything else?

_No response_

GuiEpi:
close #2154

Maybe I'm missing something, but I think that's where the problem lies, not with my component.

### DIFF
--- a/src/js/intl-tel-input.ts
+++ b/src/js/intl-tel-input.ts
@@ -1618,10 +1618,12 @@ export class Iti {
       return;
     }
     this.#ui.telInput.disabled = disabled;
-    if (disabled) {
-      this.#ui.selectedCountry.setAttribute("disabled", "true");
-    } else {
-      this.#ui.selectedCountry.removeAttribute("disabled");
+    if (this.#ui.selectedCountry) {
+      if (disabled) {
+        this.#ui.selectedCountry.setAttribute("disabled", "true");
+      } else {
+        this.#ui.selectedCountry.removeAttribute("disabled");
+      }
     }
   }
 


GuiEpi:
We can also just do: `this.#ui.selectedCountry?.setAttribute("disabled", "true");`

As suggested here: https://github.com/jackocnr/intl-tel-input/issues/2154#issue-4107997865

jackocnr:
Looks good to me! Thanks Guillaume

jackocnr:
Included in v26.9.0. Thanks again.

PR #2157, opened 2026-03-27
fix: bind isValidNumber methods to preserve private field access through proxies

### LINKED ISSUE #2149: Error Uncaught TypeError: can't access private field or method in isValidNumber starting from v26.8.0
## Bug description

After upgrading **intl-tel-input** to version **26.8.0**, an exception started appearing when calling `isValidNumber()`.

The error in the browser console:
Uncaught TypeError: can’t access private field or method: object is not the right class

This issue did **not occur in previous versions** of the library and started only after upgrading to **26.8.0**.

## Expected behavior

Calling `isValidNumber()` should validate the phone number and return a boolean value without throwing an exception.

## Actual behavior

The call to `isValidNumber()` throws a runtime exception:
Uncaught TypeError: can’t access private field or method: object is not the right class

Because of this, validation logic relying on `isValidNumber()` breaks.

## Regression

This appears to be a **regression introduced in version 26.8.0**.

Downgrading to **26.7.x** resolves the issue and `isValidNumber()` works as expected.

## Environment

- Library: **intl-tel-input**
- Version with issue: **26.8.0**
- Last working version: **26.7.x**
- Framework: **Alpine.js**
- Browsers tested:
  - Chrome
  - Firefox
  - Safari

## Additional context

The issue occurs when the library is used inside a reactive component (in this case **Alpine.js**).  
Earlier versions of `intl-tel-input` worked correctly in the same environment without any changes in the component code.

Since the problem appeared immediately after upgrading to **26.8.0**, it may be related to internal changes in the library implementation in this release.

cyphercodes:
## Problem

When the <code>Iti</code> instance is wrapped in a JavaScript Proxy (e.g., by Alpine.js or other reactive frameworks), calling <code>isValidNumber()</code> throws an error:

> Uncaught TypeError: can't access private field or method: object is not the right class

This happens because private fields are tied to the original object, not the proxy.

## Solution

Bind the <code>isValidNumber()</code> and <code>isValidNumberPrecise()</code> methods in the constructor, ensuring they maintain the correct 'this' context and private field access even when called through a proxy.

## Changes

- Added method bindings in the constructor for <code>isValidNumber</code> and <code>isValidNumberPrecise</code>

## Testing

Verified that bound methods preserve private field access when accessed through a Proxy:

Fixes #2149

### DIFF
--- a/src/js/intl-tel-input.ts
+++ b/src/js/intl-tel-input.ts
@@ -103,6 +103,11 @@ export class Iti {
     this.#numerals = new Numerals(input.value);
     this.promise = this.#createInitPromises(this.#options);
 
+    //* Bind public methods to preserve private field access when instance is wrapped in a Proxy
+    //* This fixes issues with reactive frameworks (e.g., Alpine.js) that proxy the instance
+    this.isValidNumber = this.isValidNumber.bind(this);
+    this.isValidNumberPrecise = this.isValidNumberPrecise.bind(this);
+
     //* Process onlyCountries or excludeCountries array if present.
     this.#countries = processAllCountries(this.#options);
 


jackocnr:
Hmm, I appreciate that this may fix your specific issue, but (1) why just those two methods, and not all public [methods](https://intl-tel-input.com/docs/methods#instance-methods)? And (2) is this really the responsibility of this class? It feels like this should be handled by whatever library is doing the proxying, no?

cyphercodes:
@jackocnr Thanks for the review! You're right to question this. I chose those two methods because they're the ones specifically mentioned in the issue that users were having problems with when proxied. However, I understand your concern about consistency.

Regarding responsibility - the issue is that when these methods are accessed through a Proxy, the 'this' binding is lost. While the proxying library could handle this, it would need to know about the internal private fields of this class. By binding the methods in the class itself, we ensure they work correctly regardless of how they're accessed.

Would you prefer I bind all public methods for consistency? Or would you like me to explore a different approach?

jackocnr:
Sorry, but I don't want to pollute the core plugin codebase with hacks like this. I would encourage you to explore alternative solutions.

PR #2168, opened 2026-05-12
fix: allow native paste events in strictMode

### LINKED ISSUE #2167: strictMode paste handler cancels native paste events
### Prerequisites

- [x] I am using the latest version (v28.0.7) of both intl-tel-input and utils.js.
- [x] We use libphonenumber for formatting, validation, and placeholder numbers. This issue is not about libphonenumber formatting/validation/placeholder metadata.

### Integration

Vanilla JavaScript library

### Current behaviour

When `strictMode` is enabled, the core `paste` event handler calls `preventDefault()` before sanitising the pasted value and writing it back to the input.

That keeps strict-mode sanitisation working, but it also means pages using the input can fail Lighthouse's Best Practices audit: "Prevents users from pasting into input fields".

Chrome's audit checks non-readonly inputs and fails when a custom `paste` handler cancels the event: https://developer.chrome.com/docs/lighthouse/best-practices/paste-preventing-inputs/

### Expected behaviour

The native `paste` event should not be cancelled.

`strictMode` should still preserve its current behaviour after the browser performs the paste: sanitise pasted values, cap overly long values, preserve accepted numerals, update formatting/country state, and emit `strict:reject` when pasted input is stripped or rejected.

### Steps to reproduce

1. Initialise an input with `strictMode: true`.
2. Paste into the input, or run Lighthouse Best Practices against a page containing that input.
3. Observe that the `paste` event is prevented by the strict-mode handler, and Lighthouse reports "Prevents users from pasting into input fields".

### Anything else?

I can open a PR that keeps strict-mode sanitisation but moves it to the following `input` event (`inputType === "insertFromPaste"`), so the browser's native paste event remains allowed.

Poliklot:
### Summary

- Allow native `paste` events to proceed when `strictMode` is enabled.
- Store the pre-paste value/selection during `paste`, then sanitise on the following `input` event (`inputType === "insertFromPaste"`).
- Preserve existing strict-mode paste behaviour: invalid characters are stripped, overly long values are capped or rejected, accepted numeral sets are preserved, and `strict:reject` still fires where appropriate.

### Root cause

The strict-mode paste handler previously called `preventDefault()` so it could fully control the pasted value before insertion. That works functionally, but it also causes Lighthouse to report "Prevents users from pasting into input fields" because the native `paste` event is cancelled.

### Testing

- `npx vitest run tests/integration/options/strictMode.test.js tests/integration/core/easternNumerals.test.js`
- `npm run test:js`
- `npm run typecheck:main`
- `npm run lint:js`

Fixes #2167

### DIFF
--- a/packages/core/src/js/intlTelInput.ts
+++ b/packages/core/src/js/intlTelInput.ts
@@ -548,10 +548,13 @@ export class Iti {
       return;
     }
     let inputValue = this.#getTelInputValue();
+    const isPaste = e?.inputType === INPUT_TYPES.PASTE;
+    const isStrictPaste = strictMode && isPaste;
 
     //* Android workaround for handling plus when separateDialCode enabled
     if (
       this.#isAndroid &&
+      !isPaste &&
       e?.data === "+" &&
       separateDialCode &&
       allowDropdown &&
@@ -564,6 +567,7 @@ export class Iti {
     //* Android strictMode workaround: the keydown-based filter can't block these
     if (
       this.#isAndroid &&
+      !isPaste &&
       strictMode &&
       (e?.data === " " || e?.data === "-" || e?.data === ".")
     ) {
@@ -571,13 +575,11 @@ export class Iti {
       return;
     }
 
-    let didHandleStrictPaste = false;
-    if (strictMode && e?.inputType === INPUT_TYPES.PASTE) {
-      const didRejectPaste = this.#handleStrictPasteInputEvent(inputValue);
+    if (isStrictPaste) {
+      const didRejectPaste = this.#handleStrictPasteInputEvent();
       if (didRejectPaste) {
         return;
       }
-      didHandleStrictPaste = true;
       inputValue = this.#getTelInputValue();
     }
 
@@ -588,9 +590,9 @@ export class Iti {
 
     //* If user types their own formatting char (not a plus or a numeric), or they paste something, then set the override.
     const isFormattingChar =
-      !didHandleStrictPaste && e?.data && REGEX.NON_PLUS_NUMERIC.test(e.data);
-    const isPaste = e?.inputType === INPUT_TYPES.PASTE && inputValue;
-    if (isFormattingChar || (isPaste && !strictMode)) {
+      !isStrictPaste && e?.data && REGEX.NON_PLUS_NUMERIC.test(e.data);
+    const isNonStrictPaste = isPaste && inputValue && !strictMode;
+    if (isFormattingChar || isNonStrictPaste) {
       this.#userOverrideFormatting = true;
     } else if (!REGEX.NON_PLUS_NUMERIC.test(inputValue)) {
       //* If user removes all formatting chars, then reset the override.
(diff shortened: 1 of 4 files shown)


jackocnr:
Thanks for putting this together. One question: does this not cause a flash where the initially pasted text (e.g. containing letters) is added to the input, before the letters are removed?

Poliklot:
@jackocnr 

Good question. The raw pasted value does briefly exist in the input, but it is cleaned synchronously in the same `input` event before the browser gets a chance to paint another frame.

The sequence is:

1. `paste` fires, and we store the previous value/selection.
2. The browser performs the native paste.
3. The following `input` event fires with `inputType === "insertFromPaste"`.
4. We synchronously replace the value with the sanitised one.

So the unsanitised value can be observed by JS during the `input` event, but it should not be visible to the user as a painted frame.

I made a small demo here to check exactly that:
https://codesandbox.io/p/sandbox/4rmzgm

The “sync cleanup” case follows the same approach: store the paste snapshot, allow the native paste, then clean the pasted fragment synchronously in the following input event. A `requestAnimationFrame` watcher checks whether the raw value is ever seen in a rendered frame. The delayed-control case waits before cleaning, and that one does detect the raw value in a frame.

I also tested this against the real `intl-tel-input` demo locally with Chrome CPU throttling set to 20x. In that test:

- the `paste` event was not prevented;
- the native pasted value was visible to a capture-phase `input` listener;
- `intl-tel-input` cleaned it before the event completed;
- the first `requestAnimationFrame` after the paste already saw the sanitised value.

So I don't think this should produce a visible flash, even on a slowed-down device.

jackocnr:
> the unsanitised value can be observed by JS during the input event, but it should not be visible to the user as a painted frame.

Fab thanks.

I think this change is good then. Just a few things I'd like addressed before merging:

### 1. Android branches now intercept pastes (confirmed regression on Chromium)

In `#handleInputEvent`, these two Android branches run *before* the new strict-paste branch and key off `e.data`:

```ts
if (this.#isAndroid && e?.data === "+" && separateDialCode && ...) { ... }
if (this.#isAndroid && strictMode && (e?.data === " " || e?.data === "-" || e?.data === ".")) { ... }
```

In the old flow these never fired for pastes, because the paste handler called `preventDefault()` and dispatched a synthetic `input` event with no `data`. With native paste now flowing through, Chromium sets `e.data` to the pasted text on `insertFromPaste`, so on Chrome/Android a paste of `"+"`, `" "`, `"-"`, or `"."` will hit those branches — e.g. a pasted space dispatches `strict:reject` with `source: "input"` rather than `source: "paste"`, and a pasted `"+"` opens the dropdown under `separateDialCode + countrySearch`.

Fix: gate both Android branches on `e?.inputType !== INPUT_TYPES.PASTE` — they're intended for typed Android keystrokes, not pastes. Or maybe add a `const isAndroidPaste` for re-usability.

### 2. `#handleStrictPasteInputEvent` should bail when there's no snapshot

Without a snapshot we have no `pastedRaw`, no pre-paste value, and no original selection — so none of the sanitisation logic is meaningful, and on rejection `#restoreValueBeforeStrictPaste(null)` wipes the field entirely (destroying whatever the user had there). I'd just bail at the top of the function and let the normal input flow handle the value:

```ts
#handleStrictPasteInputEvent(inputValue: string): boolean {
  const pasteSnapshot = this.#strictPasteSnapshot;
  this.#strictPasteSnapshot = null;
  if (!pasteSnapshot) {
    return false;
  }
  // … rest of the function, with all the `?? …` fallbacks removed
}
```

Returning `false` lets `#handleInputEvent` continue with country-update + format-as-you-type on whatever value is in the field. Worst case: an `insertFromPaste` we couldn't track slips through unsanitised — much better than wiping the user's input. This also lets you delete all the `?? …` fallbacks and the `if (!pasteSnapshot)` branch inside `#restoreValueBeforeStrictPaste`.

### 3. Simplify `didHandleStrictPaste`

The `let didHandleStrictPaste` flip can collapse to a single `const isStrictPaste = strictMode && e?.inputType === INPUT_TYPES.PASTE;`. The rejection path early-returns, so `!isStrictPaste && …` has identical semantics in the `isFormattingChar` check.

Once these are addressed I'm happy to merge. Thanks again!

Poliklot:
@jackocnr 

Thanks, great catch.

I addressed all three points:

1. The Android input workarounds now ignore paste input events, so native pastes with `data` like `"+"`, `" "`, `"-"` or `"."` are handled by the strict-paste path instead of the Android typed-input branches.

2. `#handleStrictPasteInputEvent` now bails out when there is no paste snapshot. In that case it returns `false` and lets the normal input flow continue, rather than trying to reconstruct the paste and potentially restoring from `null`.

3. I replaced the mutable `didHandleStrictPaste` flag with `isStrictPaste`.

I also added regression coverage for:
- strict paste input with no snapshot not clearing existing content;
- Android paste of `"+"` not opening country search under `separateDialCode`;
- Android paste of invalid content being reported as `source: "paste"` rather than `source: "key"`.

I re-ran:
- `npm run lint:js`
- `npm run typecheck:main`
- `npm run test:js`

All passing.

jackocnr:
Amazing, thanks for your work on this. Released in v28.0.9.

PR #2179, opened 2026-07-02
Use code snippets for CSS variables 

ZhangYiJiang:
The `typographer` option for markdown-it converts `--` into em dashes. Unfortunately this breaks the CSS variables in https://intl-tel-input.com/docs/theming so if you try to copy them, they don't work. Wrapping them in code block fixes this 

Screenshot 2026-07-02 at 11 12 14 AM


jackocnr:
Good catch, thanks so much!

PR #2181, opened 2026-07-05
[WIP] RTL: use CSS logical properties instead of physical + [dir=rtl] overrides

johnnyshields:
## Important

This breaks support for IE 11 and earlier. It's probably worth to release this as a major version which makes the IE browser support drop clear. (Users can use earlier ITI versions if they require IE support.)
https://caniuse.com/?search=padding-inline

---

## Summary

Converts the directional CSS in `intlTelInput.css` from physical properties (and `[dir="rtl"]` overrides) to **CSS logical properties**, so RTL is handled automatically from `dir="rtl"` alone — no special-case blocks needed. Also updates the JS input-padding to use `paddingInlineStart` so it lands on the correct side in RTL.

Motivation: today RTL relies on a handful of `[dir="rtl"]` overrides plus physical `left/right`/`padding-*`/`margin-*`, and the JS hard-codes `paddingLeft`. Consumers whose input has the selected-country on the inline-start end up needing their own overrides in RTL. Logical properties make the whole widget mirror correctly from the writing direction, with less CSS.

## Changes

**`packages/core/src/css/intlTelInput.css`**
| Physical | Logical |
|---|---|
| `left` / `right` (positioning) | `inset-inline-start` / `inset-inline-end` |
| `padding-left` / `padding-right` | `padding-inline-start` / `padding-inline-end` (or `padding-inline`) |
| `margin-left` / `margin-right` | `margin-inline-start` / `margin-inline-end` |
| `padding: 0 A 0 B` | `padding-block` + `padding-inline` |

Removed the now-redundant `[dir="rtl"]` blocks: search-input padding, search icon, search clear, and country-list flag margin.

**`packages/core/src/js/core/ui.ts`** — `#updateInputPadding` sets `style.paddingInlineStart` (and captures/restores `paddingInlineStart`) instead of `paddingLeft`, matching the selected-country which is positioned at `inset-inline-start`.

## Intentionally left physical (not direction-sensitive)

- The arrow triangle (`border-right`/`border-bottom` + `rotate`) — a down-caret shouldn't mirror.
- The detached/fullscreen popup positioning — viewport pixel math from `getBoundingClientRect` and CSS anchor positioning (`anchor(left)`), which are coordinate-based, plus the full-screen overlay's `inset: 0`.
- The loading spinner's `border-right-color`.

## Notes

- Behaviour in LTR is unchanged (logical-start == left).
- Built and smoke-tested LTR + RTL; the country container, dropdown, search box, list rows, flag/dial-code spacing, and input flag-clearance padding all mirror correctly from `dir="rtl"` with no consumer overrides.
- Logical properties and `inset-*` are supported in all current evergreen browsers (and Safari 14.1+/15+).

Happy to adjust naming or split the CSS/JS if you'd prefer.


jackocnr:
Hi there, thanks for putting this together. It's not clear from the description, but it sounds like this will change the RTL layout - can you specify exactly how it would look/work in the RTL case? (e.g. position of selected flag, dropdown chevron, selected dial code, and alignment/direction of typed phone number)

Please also see a relevant discussion with a native Persian speaker here: https://github.com/jackocnr/intl-tel-input/issues/1936

jackocnr:
Hi @johnnyshields, is there any update on this, or shall I close it?

jackocnr:
Closing due to inactivity.

PR #2183, opened 2026-07-06
Docs: Clarify hiddenInputs option behavior

lmoulanier:
Hi, I just wanted to clarify the behavior of the [hiddenInputs](https://intl-tel-input.com/docs/options#hiddeninputs) option when an input with that name already exists on the form.

Thank you.


jackocnr:
Thanks for this

PR #2184, opened 2026-07-21
Expose no-assets build in exports

ZhangYiJiang:
There's a no-assets build in `dist/css/intlTelInput-no-assets.css` but it's not exposed via the `exports` field so bundlers won't import it, this just adds a new export field for it


jackocnr:
Thanks for this.

jackocnr:
Released in v29.1.3

PR #2191, opened 2026-08-21
Docs: mention countryOrder as a soft-touch alternative

### LINKED ISSUE #2190: Document how to prioritise a likely country without selecting it
### Integration

Vanilla JavaScript library

### Description

Could the documentation include a short recipe for using a contextual country hint to order the country list without making it the initial selection?

The existing API appears to support this already:

```js
const countryHint = /* optional, validated ISO alpha-2 code */;

intlTelInput(input, {
  countryOrder: countryHint ? [countryHint] : null,
});
```

With `initialCountry` and `initialCountryLookup` omitted, the input remains in its empty/globe state. When a valid hint is available, that country appears first and the omitted countries remain alphabetical.

The host application can obtain the hint from an explicit account preference, trusted server/CDN country header, or another source it already has. No additional IP-geolocation request is required.

This is a documentation proposal only; I am not proposing a new option or default behaviour.

### Use case

A country hint can be useful without being reliable enough to determine the telephone country automatically. Travellers, VPNs, stale account data, and imperfect geolocation can all make automatic selection incorrect.

Prioritising the likely country is lower-risk: a good hint reduces scrolling or searching, while a bad hint does not change the field's value or prevent the user from finding another country.

The best-practices documentation currently covers known countries through `initialCountry` and IP lookup through `initialCountryLookup`. This would document the middle case: suggest, but do not select.

I would be happy to prepare the documentation PR in whichever section the maintainers prefer.

Chris-Larkin:
Adds the short Best Practices note suggested in #2190: when a country signal is tentative, `countryOrder` can prioritise it without turning it into the selected country.

Documentation only; no code changes.

Closes #2190.


jackocnr:
Thanks Chris

PR #2194, opened 2026-08-25
Make the country list keyboard-focusable to fix a11y scrollable-region check

### LINKED ISSUE #2192: A11y: country list fails axe scrollable-region-focusable requirement when countrySearch is disabled
### Prerequisites

- [x] I am using the latest version (v29.2.3) of both intl-tel-input and utils.js.
- [x] We use libphonenumber for formatting, validation, and placeholder numbers. If my issue relates to one of these things, I confirm I have used their [test site](https://libphonenumber.appspot.com) to ensure the issue is not with them, and will provide a link to the test results page showing this. Else, I have [reported the issue](https://github.com/google/libphonenumber/blob/master/CONTRIBUTING.md) to them instead of here.

### Integration

Vanilla JavaScript library

### Current behaviour

When `countrySearch` is disabled and the country selector is open, axe DevTools reports a `scrollable-region-focusable` violation on the country list:

```text
[serious] scrollable-region-focusable:
Scrollable region must have keyboard access

Element should have focusable content
Element should be focusable
```

The affected element is:

```html
<ul
  class="iti__country-list"
  id="iti-0__country-listbox"
  role="listbox"
  aria-label="List of countries"
>
```

The list is a scrollable region, but it has no `tabindex`. Its `role="option"` children use `tabindex="-1"`, so the region contains no focusable content.

Country selection nevertheless appears keyboard-accessible because focus remains on the selected-country button and the library handles Arrow Up/Down and Enter from there. However, the actual scrollable listbox does not satisfy axe's keyboard-access requirements.

### Expected behaviour

The open country selector should not trigger axe's `scrollable-region-focusable` rule.

The scrollable country list should use an accessible focus model in which the region itself is keyboard-focusable or contains keyboard-focusable content, while preserving the existing Arrow Up/Down, Enter, and Escape behaviour.

### Steps to reproduce

Prerequisite: Install the axe DevTools browser extension.

- Chrome: https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd
- Firefox: https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/

1. Open the playground with `countrySearch` disabled:
   https://intl-tel-input.com/playground?countrySearch=false
2. Ensure the country selector is open. The playground's "Keep dropdown open" option can remain enabled.
3. Open axe DevTools and run a full-page scan.
Image
 
4. Observe a `scrollable-region-focusable` violation on `.iti__country-list`.
Image
Image

### Anything else?

Observed with:

- intl-tel-input v29.2.3
- `countrySearch: false`
- axe DevTools using axe-core 4.12
- Chrome 151
- Firefox 154

Axe rule documentation:

https://dequeuniversity.com/rules/axe/4.12/scrollable-region-focusable

The same markup is exposed through the Vue component, which is how we originally encountered it, but the listbox itself is generated by the core library. I therefore selected "Vanilla JavaScript library" as the integration.

I found the related accessibility discussion in #800, but that report concerns screen-reader announcements and general ARIA problems. I could not find an existing report specifically about the scrollable country list failing axe's `scrollable-region-focusable` rule when `countrySearch` is disabled.

raphyabak:
## Summary

Fixes #2192.

The country dropdown's scrollable `<ul role="listbox">` has no `tabindex`, and all its `<li role="option">` children use `tabindex="-1"` (they're only reachable via arrow-key navigation while focus stays on the selected-country button or search input). Since the list itself isn't focusable and has no other focusable content, axe-core's `scrollable-region-focusable` rule flags it — most visibly when `countrySearch` is disabled, since there's no search input to otherwise satisfy the check.

## Fix

Add `tabindex="0"` to the country list. This doesn't change keyboard behavior: `#bindCountrySelectorKeydownListener` already listens on the whole `#countrySelectorEl` container, not just the selected-country button, so Arrow/Enter/Escape keydowns bubbling up from the list continue to work identically whether focus is on the list itself, the button, or the search input.

I verified this live against the production playground with axe-core (not just reasoning about it), since `scrollable-region-focusable` is a rule with known false positives on some listbox/combobox patterns and I wanted to be sure this was a real, fixable violation and not one of those:

- `countrySearch=false`: violation present on `#iti-*__country-listbox` before the fix, gone after adding `tabindex="0"`. `tabindex="-1"` alone does **not** satisfy the rule (axe requires actual keyboard reachability, not just programmatic focusability).
- `countrySearch=true`: the listbox is already clean either way (the search input's `aria-activedescendant` combobox pattern already satisfies it), so this change is a no-op there — no risk of double-counting a fix that wasn't needed.
- Manually dispatched `ArrowDown` with focus placed directly on the listbox: highlight still moves correctly, confirming the existing bubbling keydown handler keeps working once the list itself can hold focus.

## Test plan

- Added a markup test asserting the list has `tabindex="0"`.
- Added a keyboard-navigation test with `countrySearch: false` that dispatches `ArrowDown` from the list element itself (rather than the search input, which is what the existing arrow-key tests use) and asserts the highlight still moves — this is the scenario the fix newly makes possible (focus landing directly on the list) and the reason a plain `tabindex="-1"` wouldn't have worked.
- Verified the regression: reverting the source change turns the new `tabindex` test from passing into failing (`null` instead of `"0"`).
- Full unit suite passes: `npx vitest run tests/unit` — 175 passing.
- `eslint` passes on the changed files.


raphyabak:
Follow-up: making the country list keyboard-focusable put it into the native tab order, which broke the existing "pressing TAB closes the dropdown and focuses the input" test — a forward Tab could land on the list itself instead of continuing to the input. Pushed a fix that explicitly moves focus to the input on a forward Tab rather than relying on native traversal, plus a couple of unit tests covering both the forward-Tab and Shift+Tab cases.

jackocnr:
Thank you for putting this together. I manually merged your first commit and realised that the issue you found and fixed in your second commit was actually just a jsdom issue, which I fixed in a different way. Thanks again.

jackocnr:
Released with v29.4.0

PR #2195, opened 2026-08-28
Added Gaelic and Maltese locales

NothingJenkins:
Adds translation for Gaelic (GA) and Maltese (MT).
Both translations were done using DeepL as I don't speak any of these languages.


jackocnr:
Thanks!

jackocnr:
Released with v29.3.0

PR #2201, opened 2026-09-17
Fix onChangeCountry firing on util load 

ZhangYiJiang:
Claude found a race condition because of util loading. Because lastEmittedCountryRef ignores initialCountry, the component thinks country has changed even though it has not. 

Unfortunately it also notes the same bug exists for the other callbacks, but its fix for those are much more complex and convoluted because in those cases, the user could also have typed something, but the initial state for states like error cannot be determined until utils have loaded. I'm not sure its hacky looking fix makes sense or if we should figure out a restructure here, so I'm just submitting this fix.


jackocnr:
Thanks

jackocnr:
Released with v29.4.0

